PR #8, opened 2015-01-27
Determine the Linux distribution logic

yannmh:
Follow the same logic as in the `install_agent.sh` installation script.

Try `lsb_release` command to determine the Linux distribution and 
- If failing try `/etc/issue` should contain the release information
- If still failing, rely on the `/etc/*-release` files

More info:
http://unix.stackexchange.com/questions/92199/how-can-i-reliably-get-the-operating-systems-name

Also see:
https://github.com/DataDog/dd-agent/pull/1322

### DIFF
--- a/package-scripts/datadog-agent/preinst
+++ b/package-scripts/datadog-agent/preinst
@@ -2,7 +2,8 @@
 
 # Just in case there are leftovers from previous installation (we are now using our own
 # supervisord)
-LINUX_DISTRIBUTION=$(grep -Eo "(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)" /etc/issue)
+KNOWN_DISTRIBUTION="(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)"
+LINUX_DISTRIBUTION=$(lsb_release -d 2>/dev/null | grep -Eo $KNOWN_DISTRIBUTION  || grep -Eo $KNOWN_DISTRIBUTION /etc/issue)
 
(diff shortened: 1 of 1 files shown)


package-scripts/datadog-agent/preinst
```diff
@@ -30,7 +31,7 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
 
   #DEBHELPER#
 
-elif [ -f /etc/redhat-release ] || [ "$LINUX_DISTRIBUTION" == "RedHat" ] || [ "$LINUX_DISTRIBUTION" == "CentOS" ] || [ "$LINUX_DISTRIBUTION" == "openSUSE" ] || [ "$LINUX_DISTRIBUTION" == "Amazon" ]; then
+elif [ -f /etc/redhat-release ] || [ -f /etc/system-release ] || [ "$LINUX_DISTRIBUTION" == "RedHat" ] || [ "$LINUX_DISTRIBUTION" == "CentOS" ] || [ "$LINUX_DISTRIBUTION" == "openSUSE" ] || [ "$LINUX_DISTRIBUTION" == "Amazon" ]; then
```
remh: Is /etc/system-release present only on Amazon Linux ?

yannmh: Not Amazon Linux only, but CentOS/Fedora based systems.

LeoCavaille:
Merging this change too after merging the same change in the agent for the 5.2 builds.

PR #9, opened 2015-02-02
Leo/cleanup

remh:


build
```diff
@@ -0,0 +1,55 @@
...
+
+####################
+# Set up defaults
+###################
+
+: ${DISTRO:?"Need to set DISTRO"}
+: ${ARCH:?"Need to set ARCH"}
+
+# Defaults for jenkins builds and local builds
+export LOG_LEVEL="debug"
```
remh: You shouldn't default to debug, i've noticed in the past that it slows the build down by a significant factor

remh:
Looks great otherwise !

build
LeoCavaille: Thanks for catching that, that was for debug purposes, the switch below was made for that. rebasing this now.

PR #11, opened 2015-02-09
[osx] Add Mac OS X build instructions

degemer:
Modify instructions for Mac build ; shouldn't affect the other builds.

Related to : 
- https://github.com/DataDog/omnibus-software/pull/8
- https://github.com/DataDog/dd-agent/pull/1361


package-scripts/datadog-agent/preinstall
```diff
@@ -0,0 +1,56 @@
...
+
+    echo '# Stop old agent'
+    $DD_COMMAND stop || true
+
+    echo '# Stop old GUI'
+    kill `ps aux | grep 'Datadog Agent.app' | grep -v grep  | awk '{ print $2 }'` || true
+
+    echo '# Save old conf'
+    mkdir -vp /tmp/{conf,checks}.d
+    rm -vf /tmp/{conf,checks}.d/* /tmp/datadog.conf
```
elafarge: Couldn't this line be a problem if someone had the (bad) idea to split some components of his checks into different files stored in a subfolder ?

degemer: It's for Mac, which is (will be) more a test platform than anything else. It's **really** an edge case; moreover it's a precaution which was needed when the agent was contained in the app, I'm not even sure it's useful anymore.
So I'd say let's do it FB-way, _break it then fix it_. If someone reports this issue, we'll update it.

degemer:
Merged in https://github.com/DataDog/dd-agent-omnibus/pull/35.

PR #16, opened 2015-04-02
Installation failing on Jessie when datadog.conf doesn't exist.

elafarge:
We needed to add the set +e so that the deb post install script doesn't block th...e whole installation process if it fails because of a non-existing /etc/dd-agent/datadog.conf file.

My main concern is "why do we need to restart our agent here ??". And why was it working before as well because in the end it wasn't really related to systemd ?

### DIFF
--- a/package-scripts/datadog-agent/postinst
+++ b/package-scripts/datadog-agent/postinst
@@ -34,7 +34,6 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
           update-rc.d datadog-agent defaults
           adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
           usermod -d /opt/datadog-agent dd-agent
-          set +e
       ;;
       abort-upgrade|abort-remove|abort-deconfigure)
       ;;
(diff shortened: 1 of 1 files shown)


remh:
Thanks a lot @Sparadox ! 

I think it would make more sense to actually move the:
```bash
set +e
```
located on line 37 instead of adding a new one.

Also i'm not convinced that we need the 
```
set -e
```

elafarge:
Looking at the bottom of the script, I don't think we do as well :D

elafarge:
The new approach makes the lines at the end of the file (from RETVAL=$?) useless since in case an error is triggered, the RETVAL is assigned the return value of echo, which is always zero. On the other hand if the agent successfully restarts, RETVAL is also assigned a null value. 

In short, I think we can delete the lines 66-75. Does everybody agree with that ?

package-scripts/datadog-agent/postinst
```diff
@@ -57,10 +57,11 @@ if which chkconfig >/dev/null 2>&1; then
   chkconfig --add datadog-agent
 fi
 
+MISSING_CONF_MSG="\033[34m\n* No /etc/dd-agent/datadog.conf file has been found.\033[0m"
 if which invoke-rc.d >/dev/null 2>&1; then
-    invoke-rc.d datadog-agent restart
+    invoke-rc.d datadog-agent restart || echo "$MISSING_CONF_MSG"
 else
-    /etc/init.d/datadog-agent restart
+    /etc/init.d/datadog-agent restart || echo "$MISSING_CONF_MSG"
```
LeoCavaille: When using the `||` operator, you swallow all non-zero exit codes. In the agent init script, the "missing config" case is something that is handled differently see https://github.com/DataDog/dd-agent/blob/master/packaging/debian/datadog-agent.init#L71-L75
In case the restart failed because of some other reason and the init script exits with a non-zero and `!=3` error code, you'll still output the `MISSING_CONF_MSG` and exit gracefully ; I think that's not what's expected.

LeoCavaille: Also when using escaped chars in a string you shouldn't use `echo`, but rather `printf`(you could theoretically use `echo -e` but it's not POSIX-compliant...).

elafarge:
Thanks for mentionning that, I updated the scripts.

This being said, I think I finally spotted where the error was coming from... 

It seems that on Jessie, the invoke-rc.d line actually calls systemctl restart datadog-agent . The problem is that in case of a failure in our init script (typically if the config file doesn't exit, it fails with 3 as an error code), the exit code is not passed to the systemctl exit code (which becomes 1) and thus, it's impossible to check if the error is due to a missing config file or something dirtier.

I see two solutions : wrapping our agent in a systemd compliant service (which we will probably have to do at some point) or checking if systemd is installed on the machine and if so, restart only if the service is started (which can be done in one command with systemd). I'll work on the second solution to have something ready before Wednesday but keeping the first one in mind would be good.

elafarge:
I just finished implementing and testing the second solution.

package-scripts/datadog-agent/postinst
```diff
@@ -53,23 +52,28 @@ chown -R dd-agent:root ${LOG_DIR}
 chown root:root /etc/init.d/datadog-agent
 chown -R root:root /opt/datadog-agent
 
-if which chkconfig >/dev/null 2>&1; then
-  chkconfig --add datadog-agent
-fi
-
-if which invoke-rc.d >/dev/null 2>&1; then
-    invoke-rc.d datadog-agent restart
+if which systemctl >/dev/null 2>&1; then
```
LeoCavaille: You should use `command -v` instead which is defined in the POSIX standards.

LeoCavaille:
I like the `try-restart` approach but it's still not completely equivalent to the `invoke-rc.d` behavior.

Examples:

1. the agent was stopped before the upgrade
    * before: it would be restarted by the upgrade
    * after: no

2. the agent had a buggy config and was stopped
    * before: the upgrade would fail, because the init script would try to start the agent and exit non-zero
    * after: the upgrade would silently succeed

These are I guess edge cases, but they at least try to ensure that the agent is running at the end of the install. @remh do you want to chime in?

PR #17, opened 2015-04-08
Remove `Provides` from package metadata

LeoCavaille:
This is to avoid issues with yum which uninstalls `datadog-agent` when
running `yum remove datadog-agent-base` because `datadog-agent` provides
`datadog-agent-base`

### DIFF
--- a/config/projects/datadog-agent.rb
+++ b/config/projects/datadog-agent.rb
@@ -21,10 +21,12 @@ description "Datadog Monitoring Agent
 "
 platform_in_iteration false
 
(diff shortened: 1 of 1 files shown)


remh:
:+1:

LeoCavaille:
Yes!
```
[root@puppetmaster ~]# yum info datadog-agent | grep Version
Version     : 5.3.0+git.280.9a4f66d
[root@puppetmaster ~]# yum remove datadog-agent-base
Loaded plugins: fastestmirror
No Match for argument: datadog-agent-base
No Packages marked for removal
```

PR #18, opened 2015-04-15
[systemd] made the agent systemd compliant

elafarge:
Everything is in the title. Careful there's also a modification on
install_agent.sh in the dd-agent project.

### DIFF
--- a/package-scripts/datadog-agent/postinst
+++ b/package-scripts/datadog-agent/postinst
@@ -32,7 +32,7 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
     case "$1" in
         configure)
             update-rc.d datadog-agent defaults
-            update-rc.d datadog-agent enable
+            update-rc.d datadog-agent enable >/dev/null 2>&1
             adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
             usermod -d /opt/datadog-agent dd-agent
             set +e
--- a/package-scripts/datadog-agent/postinst
+++ b/package-scripts/datadog-agent/postinst
@@ -7,16 +7,16 @@ LINUX_DISTRIBUTION=$(grep -Eo "(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)" /e
 
 error_exit()
 {
-  echo "${PROGNAME}: ${1:-"Unknown Error"}" 1>&2
-  exit 1
+    echo "${PROGNAME}: ${1:-"Unknown Error"}" 1>&2
+    exit 1
 }
 
 mkdir -p ${LOG_DIR} || error_exit "Cannot create ${LOG_DIR}!"
 
 # If we are inside the Docker container, do nothing
 if [ -n "$DOCKER_DD_AGENT" ]; then
-  echo "Installation from docker-dd-agent, nothing to do in postinst"
-  exit 0
+    echo "Installation from docker-dd-agent, nothing to do in postinst"
+    exit 0
 fi
 
 # Since we now package our own supervisor config, we no longer want
@@ -28,24 +28,23 @@ if [ -f /etc/supervisor/conf.d/ddagent.conf ]; then
 fi
 
 if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LINUX_DISTRIBUTION" == "Ubuntu" ]; then
-  set -e
-  case "$1" in
-      configure)
-          update-rc.d datadog-agent defaults
-          adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
-          usermod -d /opt/datadog-agent dd-agent
-          set +e
-      ;;
-      abort-upgrade|abort-remove|abort-deconfigure)
-      ;;
+    set -e
+    case "$1" in
+        configure)
+            update-rc.d datadog-agent defaults
+            update-rc.d datadog-agent enable
+            adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
+            usermod -d /opt/datadog-agent dd-agent
+            set +e
+        ;;
+        abort-upgrade|abort-remove|abort-deconfigure)
+        ;;
 
-      *)
-      ;;
-  esac
-  #DEBHELPER#
+        *)
+        ;;
+    esac
+    #DEBHELPER#
 fi
-
-chmod 755 /etc/init.d/datadog-agent
 
 # Set proper rights to the dd-agent user
 chown -R dd-agent:root ${CONFIG_DIR}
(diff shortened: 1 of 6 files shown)


package-scripts/datadog-agent/postinst
```diff
@@ -31,7 +33,11 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
   set -e
   case "$1" in
       configure)
-          update-rc.d datadog-agent defaults
+          if which systemd >/dev/null 2>&1; then
+	    ln -s /lib/systemd/system/datadog-agent.service /etc/systemd/system/datadog-agent.service
+	  else
+	    update-rc.d datadog-agent defaults
+	  fi
```
LeoCavaille: nitpick: the spacing looks a bit wonky here. Can we try to use 4 spaces everywhere?

package-scripts/datadog-agent/postinst
```diff
@@ -31,7 +33,11 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
   set -e
   case "$1" in
       configure)
-          update-rc.d datadog-agent defaults
+          if which systemd >/dev/null 2>&1; then
```
LeoCavaille: Use `command -v` instead of `which` _which_ is not in the POSIX standards..

package-scripts/datadog-agent/postinst
```diff
@@ -31,7 +33,11 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
...
+	    ln -s /lib/systemd/system/datadog-agent.service /etc/systemd/system/datadog-agent.service
```
LeoCavaille: This command will fail if the target already exists.

LeoCavaille: Can't we use a systemd enable command for that instead btw?

package-scripts/datadog-agent/postinst
```diff
@@ -5,6 +5,8 @@ LOG_DIR=/var/log/datadog
 
 LINUX_DISTRIBUTION=$(grep -Eo "(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)" /etc/issue)
 
+chmod 755 /etc/init.d/datadog-agent
```
LeoCavaille: Can you add a comment here, so that we remember why this instruction needs to be run first.

package-scripts/datadog-agent/postinst
```diff
@@ -57,20 +62,24 @@ if which chkconfig >/dev/null 2>&1; then
   chkconfig --add datadog-agent
 fi
 
-if which invoke-rc.d >/dev/null 2>&1; then
-    invoke-rc.d datadog-agent restart
+if which systemd >/dev/null 2>&1; then
```
LeoCavaille: Same for `which`

package-scripts/datadog-agent/postinst
```diff
@@ -57,20 +62,24 @@ if which chkconfig >/dev/null 2>&1; then
...
-RETVAL=$?
-if [ $RETVAL -ne 0 ]; then
-    if [ $RETVAL -eq 3 ]; then
-        # No datadog.conf file is present. The user is probably following
-        # the step-by-step instructions and will add the config file next.
-        exit 0
-    else
-        exit $RETVAL
-    fi
+  if which invoke-rc.d >/dev/null 2>&1; then
```
LeoCavaille: This should probably be elif, but again isn't `invoke-rc.d` actually calling `systemctl` when you run systemd?

package-scripts/datadog-agent/postrm
```diff
@@ -4,12 +4,17 @@ LINUX_DISTRIBUTION=$(grep -Eo "(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)" /e
...
+	    systemctl remove datadog-agent
+	else
+	    update-rc.d datadog-agent remove
+        fi
         deluser dd-agent
         rm -rf /opt/datadog-agent
         rm -rf /var/log/datadog
         rm -rf /etc/dd-agent
         rm -rf /var/log/datadog
+	rm -rf /lib/systemd/system/datadog-agent.service
```
LeoCavaille: nitpick, but should probably lose the recursive flag

package-scripts/datadog-agent/postrm
```diff
@@ -4,12 +4,17 @@ LINUX_DISTRIBUTION=$(grep -Eo "(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)" /e
 if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LINUX_DISTRIBUTION" == "Ubuntu" ]; then
     set -e
     if [ "$1" = purge ]; then
-        update-rc.d datadog-agent remove
+        if which systemd >/dev/null 2>&1; then
+	    systemctl remove datadog-agent
+	else
+	    update-rc.d datadog-agent remove
+        fi
```
LeoCavaille: Same Q here.

package-scripts/datadog-agent/postinst
elafarge: Yes it is, but the problem is the return value. Systemctl seems to swallow the exit code of the service it is running. Though I found a hack (pull coming soon, I'm just finishing reviewing your review)

package-scripts/datadog-agent/postinst
elafarge: Yes we can. Weirdly though it may be, the update-rc.d call isn't equivalent to systemctl enable (no symlink exists under /etc/systemd/system after the call).

package-scripts/datadog-agent/postinst
elafarge: Done

package-scripts/datadog-agent/postinst
elafarge: y

package-scripts/datadog-agent/postrm
elafarge: This line has been removed anyway since this file is removed without our help in postrm

package-scripts/datadog-agent/postinst
elafarge: Note : after all it seems to work. See new commits.

package-scripts/datadog-agent/postrm
```diff
@@ -3,8 +3,15 @@ LINUX_DISTRIBUTION=$(grep -Eo "(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)" /e
 
 if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LINUX_DISTRIBUTION" == "Ubuntu" ]; then
     set -e
+    
+    # Removing the service form startup (since it's not there anymore)
+    if command -v systemd >/dev/null 2>&1; then
+        systemctl disable datadog-agent
```
LeoCavaille: Is there an `update-rc.d disable` that would do the opposite of `update-rc.d enable` you used in the postinst script?

package-scripts/datadog-agent/postinst
```diff
@@ -28,41 +28,45 @@ if [ -f /etc/supervisor/conf.d/ddagent.conf ]; then
...
+if command -v chkconfig >/dev/null 2>&1; then
+    chkconfig --add datadog-agent
 fi
 
-if which invoke-rc.d >/dev/null 2>&1; then
-    invoke-rc.d datadog-agent restart
-else
-    /etc/init.d/datadog-agent restart
-fi
+/etc/init.d/datadog-agent configcheck
```
LeoCavaille: Maybe add a comment here to explain the different status codes and why we skip restart in those cases?

package-scripts/datadog-agent/postrm
elafarge: Indeed there is thanks. The behaviour on Jessie is the expected one, checking on a non systemd distro as soon as my build is complete.

package-scripts/datadog-agent/postinst
elafarge: Done

package-scripts/datadog-agent/postrm
elafarge: I had to move the disable call to the prerm script. It seems that if we call disable and the service file doesn't exit, it tries to disable it as if it was a SysVInit service and fails.

LeoCavaille:
Looks like it's good to go too now. Thanks :+1:

PR #20, opened 2015-04-27
First pass at dockerized builds

remh:


README.md
```diff
@@ -5,29 +5,36 @@ This is an [Omnibus](https://github.com/opscode/omnibus) project to build the Da
 
 It's using a [fork](https://github.com/DataDog/omnibus-ruby/compare/chef:v3.2.2...custom3.2.2-2) of the official 3.2.2 release of the Omnibus project.
 
-## Use omnibus locally
+Builds are run in docker containers with Circleci.
+See:
+* https://github.com/DataDog/docker-dd-agent-build-deb-i386
+* https://github.com/DataDog/docker-dd-agent-build-rpm-i386
+* https://github.com/DataDog/docker-dd-agent-build-deb-x64
+* https://github.com/DataDog/docker-dd-agent-build-rpm-x64
```
LeoCavaille: This makes me a bit sad... no way around it :/ ?

omnibus_build.sh
```diff
@@ -1,27 +1,19 @@
...
-su vagrant -c "bundle install --binstubs"
-su vagrant -c "bundle update" # Make sure to update to the latest version of omnibus-software
-su vagrant -c "./bin/omnibus build -l=$LOG_LEVEL $PROJECT_NAME"
-
-# TODO: add rpm signing
-#if [ #{ENV['PKG_TYPE']} == "rpm" ] && [ #{ENV['GPG_PASSPHRASE']} ] && [ #{ENV['GPG_KEY_NAME']} ]; then
-#    chmod +x rpm-sign
-#    su vagrant -c "./rpm-sign #{ENV['GPG_KEY_NAME']} #{ENV['GPG_PASSPHRASE']} pkg/#{project_name}-#{ENV['AGENT_VERSION']}-#{ENV['BUILD_NUMBER']}.*.rpm"
-#fi
+/bin/bash -l -c "git pull"
```
LeoCavaille: Why do we need this? The build script shouldn't modify my working files.

omnibus_build.sh
```diff
@@ -1,27 +1,19 @@
...
+/bin/bash -l -c "rvm use 2.2.2"
+/bin/bash -l -c "bundle update" # Make sure to update to the latest version of omnibus-software
+/bin/bash -l -c "bin/omnibus build -l=$LOG_LEVEL $PROJECT_NAME"
```
LeoCavaille: Can't we just source the rvm script once and avoid to launch multiple new shells from scratch?

LeoCavaille:
Nice work, :zap: fast builds coming soon

README.md
remh: Unfortunately it's the only way to enable parallelism in circle ci jobs.

omnibus_build.sh
remh: It was to ensure to get the latest version of dd-agent-omnibus without having to recreate the docker images all the time which was long and a bottleneck for development. But dd-agent-omnibus should be more "stable" now, so we could get rid of it and trigger new docker builds when dd-agent-omnibus is updated. 

What do you think?

PR #23, opened 2015-05-13
delete all .pyc files (prerm and preinst)

degemer:
We can now safely remove files. :tada:

### DIFF
--- a/package-scripts/datadog-agent/preinst
+++ b/package-scripts/datadog-agent/preinst
@@ -29,6 +29,13 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
       rm /etc/dd-agent/supervisor_ddagent.conf
   fi
 
+  # Delete .pyc files
+  # FIXME: it shouldn't be done there, but only in prerm (see 6.6
+  # of https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html)
+  # It is also here because version < 5.4 didn't delete .pyc,
+  # so we need to be sure to clean them here (if a file is deleted for instance)
+  find /opt/datadog-agent -name '*.py[co]' -delete
+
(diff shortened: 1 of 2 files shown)


package-scripts/datadog-agent/preinst
```diff
@@ -29,6 +29,13 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
       rm /etc/dd-agent/supervisor_ddagent.conf
   fi
 
+  # Delete .pyc files
+  # FIXME: it shouldn't be done there, but only in prerm (see 6.6
+  # of https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html)
+  # It is also here because version < 5.4 didn't delete .pyc,
+  # so we need to be sure to clean them here (if a file is deleted for instance)
+  find /opt/datadog-agent/agent -name '*.pyc' -delete
```
remh: We should also remove the ones in
```
/opt/datadog-agent/embedded/lib/python2.7/
```
i think

remh: and the pyo files too, right ?

degemer:
Updated, let's delete all `.pyc` and `.pyo` of `/opt/datadog-agent`.

remh:
So i don't think we will ever be able to Do The Right Thing i.e. cleaning that in the prerm script as it wouldn't work for people upgrading from versions < 5.4.0.

Based on that, i don't think we should duplicate this code and we should just keep it in the preinst script.

@LeoCavaille what do you think ?

degemer:
Maybe for 7.x ? I mean it's really for the long term, right now it "costs" ~3s. (I ran this `find` command with no `.pyc` or `.pyo` files).

LeoCavaille:
We've discussed it, looks like a good solution for what we need now. Merging.

PR #24, opened 2015-05-13
[debian] do not stop the agent in preinst

degemer:
As it is already stopped in `prerm`, see 6.6 of
https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html

It avoids stopping twice the agent:
screen shot 2015-10-20 at 15 47 34

Otherwise, if before a specific version we weren't stopping the agent in `prerm`, this PR should not be merged. I only searched back to the first commit (5.0 version) of this repo which contains the `stop` in `prerm`: https://github.com/DataDog/dd-agent-omnibus/blob/425bfb88bbd994a0ef05bc0cc10d0ceb9eff61d9/package-scripts/datadog-agent/prerm#L5-L9


remh:
Closing as this is not an issue.

PR #26, opened 2015-05-13
Do NOT fail if /opt/datadog-agent does not exist

degemer:
Bad regression: this new `find` command fails if there is no
`/opt/datadog-agent`, which is the case for a fresh install.


package-scripts/datadog-agent/preinst
```diff
@@ -34,7 +34,7 @@ if [ -f /etc/debian_version ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$LI
   # of https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html)
   # It is also here because version < 5.4 didn't delete .pyc,
   # so we need to be sure to clean them here (if a file is deleted for instance)
-  find /opt/datadog-agent -name '*.py[co]' -delete
+  find /opt/datadog-agent -name '*.py[co]' -delete || true
```
LeoCavaille: also you could use `-type f` so that `find` only matches the pattern against files.

PR #29, opened 2015-06-10
[jmxfetch] removes JMXFetch status files

yannmh:
Upgrading from Datadog Agent 5.3.x to 5.4.0, JMXFetch is transiting from '
dd-agent' to 'root user. JMXFetch status files need to be deleted, to avoid
permission issues.

### DIFF
--- a/package-scripts/datadog-agent/postinst
+++ b/package-scripts/datadog-agent/postinst
@@ -3,6 +3,8 @@
 CONFIG_DIR=/etc/dd-agent
 LOG_DIR=/var/log/datadog
 RUN_DIR=/opt/datadog-agent/run
+JMX_STATUS_PYTHON=/tmp/jmx_status_python.yaml
+JMX_STATUS=/tmp/jmx_status.yaml
 
 LINUX_DISTRIBUTION=$(grep -Eo "(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)" /etc/issue)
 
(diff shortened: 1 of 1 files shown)


package-scripts/datadog-agent/postinst
```diff
@@ -3,6 +3,8 @@
 CONFIG_DIR=/etc/dd-agent
 LOG_DIR=/var/log/datadog
 RUN_DIR=/opt/datadog-agent/run
+JMX_STATUS_PYTHON=/tmp/jmx_status_python.yaml
+JMX_STATUS=/tmp/jmx_status.yaml
```
degemer: The only issue I see is for customers with some weird configuration about their temp directory. (since we're calling ` tempfile.gettempdir()`: https://github.com/DataDog/dd-agent/blob/6eb139010510389c4cf97a1598a93152e8320d2a/config.py#L898)
I remember seing a customer with specific `/tmp` directories for each user. (something like `/tmp/{id}`)
Not sure if there is a correct solution for it though.

yannmh: You're right, hopefully they get deleted at some point.

elafarge: On Linux systems, temporary folders are supposed to get deleted at every `shutdown`, but restarting might be a problem for some of our users, wouldn't it ? 

We could actually call a small python script to retrieve the right folder just like we're doing it here (https://github.com/DataDog/dd-agent/blob/master/packaging/datadog-agent/source/install_agent.sh#L85)... it requires on working Python installation though... 

Another alternative would be to use the `/opt/datadog-agent-embedded/bin/python` if it's there... if not the Agent is installed so we don't care. What do you think @degemer @yannmh ?

degemer: I'm not sure it would work, because this script is run as `root`, and the agent as `dd-agent`, so in this specific case `tempfile.gettempdir()` won't work. And even if we use `sudo` or `su`, some env variables won't be there...
I'm not sure it's worth spending too much time on it, it's really a small fraction of users. Maybe just warn support about it ?

degemer:
Solved by moving all status files to the internal `run` directory.

PR #32, opened 2015-06-22
Etienne/omnibus 4 migration

elafarge:
Made our `dd-agent-omnibus` build omnibus-4-compatible (for Linux only, OSX is coming in a separate branch  that has been created from this one). We had to comply with some DSL changes such as the removal of `always_build` due to the introduction of `git_cache`, the use of Omnibus file-related DSL such as `copy` instead of plain Unix commands, RPM signing features...

IMPORTANT NOTICE: when `omnibus-ruby` and `omnibus-software` will be merged to `master`, it's essential not to forget to update the Gemfile here. Merging the Omnibus 4 branches on these two projects before merging this one would be a better idea in my opinion.


Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-gem 'omnibus', git: 'git://github.com/datadog/omnibus-ruby.git', branch: 'custom3.2.2-2'
-gem 'omnibus-software', git: 'git://github.com/datadog/omnibus-software.git', branch: 'master'
+gem 'omnibus', :git => 'git://github.com/datadog/omnibus-ruby.git', :branch => 'etienne/ship-licenses-and-sources'
+gem 'omnibus-software', :git => 'git://github.com/datadog/omnibus-software.git', :branch => 'etienne/omnibus-4-migration'
```
miketheman: Are we building on something nonRuby 1.9 and above? I understand the repo changes, but the hashrocket style should be replaced with the previously existing `git: 'git...`.
Was something not working with the newer style?

config/projects/datadog-agent.rb
```diff
@@ -1,5 +1,5 @@
 name 'datadog-agent'
-maintainer 'Datadog Packages <package@datadoghq.com>'
+maintainer 'Etienne Omnitests <etienne.lafarge@datadoghq.com>'
```
miketheman: This gets populated into every package we deploy and will show up in things like apt-cache info - are you sure you should change this?

config/projects/datadog-agent.rb
```diff
@@ -18,23 +18,40 @@
  forwards metrics from your applications as well as system services.
  .
  See http://www.datadoghq.com/ for more information
-"
-platform_in_iteration false
+'
+
+# .deb specific flags
+package :deb do
+  vendor 'DataDog <info@datadoghq.com>'
```
miketheman: Dan we please use `Datadog` instead of `DataDog` wherever possible?

config/software/datadog-agent.rb
```diff
@@ -1,60 +1,58 @@
 name 'datadog-agent'
 
-local_agent_repo = ENV['LOCAL_AGENT_REPO']
+local_agent_repo = ENV["LOCAL_AGENT_REPO"]
```
miketheman: why the stylistic change? Single quotes for non-interpolated strings actually have a performance benefit.

Gemfile
elafarge: Nope, that's just me not being up to date with the newer ruby Syntax :) I'll fix that when we re-switching the branches we use to `master` (which we will do at least for `omnibus-software` I think) :)

config/projects/datadog-agent.rb
elafarge: Sure

Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-gem 'omnibus', git: 'git://github.com/datadog/omnibus-ruby.git', branch: 'custom3.2.2-2'
-gem 'omnibus-software', git: 'git://github.com/datadog/omnibus-software.git', branch: 'master'
+gem 'omnibus', git: 'git://github.com/datadog/omnibus-ruby.git', branch: 'etienne/ship-licenses-and-sources'
```
remh: Can you set the final names of the branches we will be using instead ?

config/projects/datadog-agent.rb
remh: Let's keep using package@datadoghq.com

config/projects/datadog-agent.rb
```diff
@@ -18,23 +18,40 @@
  forwards metrics from your applications as well as system services.
  .
  See http://www.datadoghq.com/ for more information
-"
-platform_in_iteration false
+'
+
+# .deb specific flags
+package :deb do
+  vendor 'Datadog <info@datadoghq.com>'
```
remh: Why is that an environment specific block  and duplicated in the RPM one ?

omnibus_build.sh
```diff
@@ -22,9 +22,20 @@ rm -rf /opt/$PROJECT_NAME/*
...
 
+# If an RPM_SIGNING_PASSPHRASE has been passed, let's import the signing key
+if [ -n "$RPM_SIGNING_PASSPHRASE" ]; then
+  gpg --import /keys/RPM-SIGNING-KEY.private
+fi
+
+# Last but not least, let's make sure that we rebuild the agent everytime because
+# the extra package files are destroyed when the build container stops (we have
+# to tweak omnibus-git-cache directly for that). Same for gohai.
+git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-agent`
```
remh: datadog-agent and datadog-gohai should have the always_build flag set to true, It's not enough ?

elafarge: Not with the new git-cache feature Omnibus4 is using. The `always_build` flag simply has no effect so I had to manually delete the tags. (Which is what is advised in Omnibus 4 cache-specific documentation page : https://github.com/chef/omnibus/blob/master/docs/Build%20Cache.md )

Gemfile
elafarge: Do you mind if I do that only after merging my changes to the final branches in question ?

config/projects/datadog-agent.rb
elafarge: Yep I changed the user to be able to use my Test key on Travis. I reverted to the old value :)

config/projects/datadog-agent.rb
elafarge: I have to duplicate it since vendor is specific to the `package` DSL and not the `project` one. In addition to that, this DSL is defined for DEB and RPM packagers but not for the other ones (OSX, Windows...)

PR #33, opened 2015-07-10
[deps] use omnibus-software agent release branch

yannmh:
A release branch is forked from master on omnibus-software repository
for each major and minor release. Use it instead of master.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 gem 'omnibus', git: 'git://github.com/datadog/omnibus-ruby.git', branch: 'custom3.2.2-2'
-gem 'omnibus-software', git: 'git://github.com/datadog/omnibus-software.git', branch: 'master'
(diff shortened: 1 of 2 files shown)


degemer:
:+1: , but for now the `5.4.3` tag is on master, and `5.4.3` commits are not on the `5.4.x` branch.
Could you also add a default to `master` in https://github.com/DataDog/dd-agent-omnibus/blob/master/omnibus_build.sh#L12 ? This way it won't break anything, and you'll be able to update easily the Jenkins job.

yannmh:
Adding a default `master` value to the branch. We should cherry-pick the changes we made from master to `5.4.x` branch like we do for `dd-agent`.

remh:
Looks good to me.

Don't forget to recreate all the docker images (github.com/DataDog/docker-dd-agent-build* ) if you want this change to be published.

yannmh:
Will do. Thanks!

PR #38, opened 2015-07-28
Ship all YAML check config examples automatically

elafarge:
Instead of maintaining a list of the examples files to ship, we add them
all automatically not to forget. Ok that's what was done before... but
since `extra_package_file` doesn't allow directories anymore it wasn't
as obvious as it looks :)


config/projects/datadog-agent.rb
```diff
@@ -84,75 +84,6 @@
   # Supervisord config file for the agent
   extra_package_file "/etc/dd-agent/supervisor.conf"
 
-  # Example configuration files for the agent and the checks
-  extra_package_file "/etc/dd-agent/datadog.conf.example"
```
miketheman: Missed?

elafarge: Yep but this won't get merged :/ Even though it's part of a build block in the Software DSLs, the code adding the example files is executed before the agent is downloaded and extracted.. I'll reopen my PR against the `chef/omnibus` repository to allow folders in this DSL :/

PR #40, opened 2015-08-24
Check for dd-agent user before creating it

talwai:
Addresses https://github.com/DataDog/dd-agent/issues/1587

### DIFF
--- a/package-scripts/datadog-agent/postinst
+++ b/package-scripts/datadog-agent/postinst
@@ -39,8 +39,14 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
           configure)
               update-rc.d datadog-agent defaults
               update-rc.d datadog-agent enable >/dev/null 2>&1
-              adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
-              usermod -d /opt/datadog-agent dd-agent
+              # Only add dd-agent user if it doesn't already exist
+              id -u dd-agent >/dev/null 2>&1
+              USER_EXISTS=$?
+              if [ ! $USER_EXISTS -eq 0 ]; then
+                  echo "Creating dd-agent user"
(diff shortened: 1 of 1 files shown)


package-scripts/datadog-agent/postinst
```diff
@@ -39,8 +39,13 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
           configure)
               update-rc.d datadog-agent defaults
               update-rc.d datadog-agent enable >/dev/null 2>&1
-              adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
-              usermod -d /opt/datadog-agent dd-agent
+              # Only add dd-agent user if it doesn't already exist
+              id -u dd-agent &>/dev/null
```
elafarge: If you really want that line to ouput nothing you should change that to `id -u dd-agent >/dev/null 2>&1` :)

package-scripts/datadog-agent/postinst
```diff
@@ -39,8 +39,13 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
...
+              USER_EXISTS=$?
+              if [ ! $USER_EXISTS -eq 0 ]; then
+                  adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
```
elafarge: I know there's not that many echoes in our `postinst` script but I think the creation of a new user on the machine is worth notifying. Could you just add a little something like `echo "Creating user 'dd-agent'."`  ?

elafarge:
I tested that on both `dash` (Ubuntu) and `bash` shells, it works like a charm. Good job !

I just had two little observations about silencing the output of one command and notifying our customers that we are actually creating a new user.

After that, let's just :ship: it !

talwai:
@elafarge Thanks for the review. Implemented your suggestions, and merging now

PR #42, opened 2015-08-31
[postinst][deb] Add error flags

talwai:
Add some echo messages and the correct error flags for capturing `dd-agent` user existence


talwai:
cc @elafarge

elafarge:
Let's merge that for 5.5 then

talwai:
Adds a bugfix to #40

PR #43, opened 2015-09-04
moving CF script to omnibus repo so it is available at package upload…

irabinovitch:
This adds a short script that can be used to invalidate CloudFront caches that sit in front of our repos when we publish new packages.


elafarge:
I updated the dockerfile to actually ship these new scripts in our Docker image. @remh I'm not completely sure about where the actual image is though. Once reviewed could we check together that it got rebuilt so that @irabinovitch can add a call to the cache invalidation scripts in our jobs ?

irabinovitch:
I'm not 100% sure this fits into dd-agent-omnibus best, but most of the other projects were a bit more awkward to pull into the build pipeline and release jobs.  Happy to move elsewhere if needed.

elafarge:
Closing, we now use a separate repository (private).

PR #45, opened 2015-09-10
Make the Omnibus build Windows compatible

elafarge:
Finally ! Here it is. First of all it has a... Batch script (yes
Batch...) to launch the build: it basically does the same as
`omnibus_build.sh` with some slight variations due to the way batches
behaves when calling subscripts (Basically if a subscripts calls `exit`
it will make the parent script exit too by default...).

Basically the `project` DSL have just been updated with some extra
definitions related to the MSI package (only) and a couple of variables
to pass to Wix... nothing too fancy in there, it just sticks to what's
written in the Omnibus documentation.

The software definitions have been updated with a couple of command to
package everything we need, build our GUI and service executables with
py2exe and ship it all.

Last but definitely not least, the default `source.wxs` file
auto-generated by Omnibus couldn't be used as such and need some heavy
modifications. Adding more features to tune it through omnibus DSLs in
the `chef/omnibus` project would be great to do... later. Note that we
also had to tune our fork of this project to actually achieve what we
wanted to do. The way Wix is used in Omnibus4 is still poorly
customizable and very rigid, but that's definitely gonna change at some
point :)

TODO (and to rebase):
- update the Gemfile before merging to use the
master branches of `omnibus-software` and `omnibus-ruby`.
- update the MSI assets with Datadog specific ones
- update the License.rtf file... it displays pretty badly in the MSI
  installer


config/projects/datadog-agent.rb
```diff
@@ -1,9 +1,18 @@
...
+if windows?
+  # Windows doesn't want our e-mail address :(
+  maintainer 'Datadog'
+else
+  maintainer 'Datadog Packages <package@datadoghq.com>'
+end
 homepage 'http://www.datadoghq.com'
-install_dir '/opt/datadog-agent'
+if ohai['platform'] == "windows"
+  install_dir "C:/Agent"
```
remh: Can you add a comment to indicate how it's used ?

config/projects/datadog-agent.rb
```diff
@@ -57,6 +66,19 @@
   pkg_position '10, 10'
 end
 
+# Windows .msi specific flags
+package :msi do
+  # For a consistent package management, please NEVER change this code
+  upgrade_code '9E1DEDC4-DE86-4714-8325-BEF9B08E34A8'
```
remh: It's not the right code

remh:
Closing this PR in favor of this branch that is still in WIP:
https://github.com/DataDog/dd-agent-omnibus/tree/quentin/omnibus-5.0

PR #46, opened 2015-09-14
[postinst] Set dd-agent as owner of the run/ directory

olivielpeau:
Was previously set correctly (see https://github.com/DataDog/dd-agent-omnibus/blob/5.4.x/package-scripts/datadog-agent/postinst#L55) but got removed by mistake, which made dd-agent unable to write pid and jmx status files in the run/ directory.


elafarge:
Thanks a lot for fixing that @olivielpeau .

PR #48, opened 2015-09-16
[osx] delete old agent files

yannmh:
On OS X, Datadog agent does not provide any uninstallation method. Thus,
agent files are not removed before an agent upgrade.
When a file is removed, renamed or moved, the old file remains and may
likely conflict with the new ones.

Delete the entire $INSTALL_DIR/agent directory, not only *.pyc, *.pyo
files.


elafarge:
Gogogo

PR #49, opened 2015-09-17
Fix dd-agent CLI on Linux

elafarge:
The `/usr/bin{dd-agent,dogstatsd,dd-forwarder}` commands haven't worked
anymore since... a big while. I suspect that this is due to a slight
variation in the behaviour of the `extra_package_file` between Omnibus 3
and Omnibus 4 but I didn't really have time to dig more into it.

This fix just moves the creation of these files to the `postinst` script
which fixes the issue in a rather simple way. These same files also get
deleted in the `postrm` script so that these commands simply don't exist
anymore after removing the agent.


talwai:
:+1:

PR #51, opened 2015-09-18
[linux] Fixed dd-agent CLI: Part II

elafarge:
The order in with package scripts get executed during updates with YUM
led to an unexpected behaviour. After a couple of minutes of
facepalming, a decent fix has been implemented. There's also a README in
the `package-scripts/datadog-agent` folder that details a bit more how
updates are handled by APT and YUM, and there's a couple of rather
important differences to pay attention to when updating these scripts :)

### DIFF
--- a/config/projects/datadog-agent.rb
+++ b/config/projects/datadog-agent.rb
@@ -88,14 +88,19 @@ if linux?
   end
 
   # Supervisord config file for the agent
-  extra_package_file "/etc/dd-agent/supervisor.conf"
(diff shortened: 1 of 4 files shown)


elafarge:
@yannmh This completes this other commit (already merged): https://github.com/DataDog/dd-agent-omnibus/commit/9b9bdcf175567be6995e37a9fb8ff0cdacb3a6b6 . Both have to be cherry-picked if we want to release a fix for that in 5.5.1 :)

yannmh:
Looks great. Thanks @elafarge !

PR #53, opened 2015-10-22
[preinst] rename log dir to correct one

degemer:


remh:
I was wondering why i had this /var/log/datad in my machine.. Thanks!

PR #57, opened 2015-12-31
Fix discrepancies on .pyc files in the deb and rpm packages

olivielpeau:
By compiling py files of the embedded python at the end of the build, we ensure that the .pyc files that are packaged are valid for their corresponding py files. They will not need to be regenerated at runtime by python and will not show up as discrepancies in the output of package verifiers.

Needs https://github.com/DataDog/omnibus-ruby/pull/12 to work.

For reference, another approach to the issue could be to:
- remove the pyc files altogether from the package
- add a step of compilation of the py files in the `postinst` script
- remove the pyc files in the `prerm` script only when the package is being removed (i.e. not upgraded)


remh:
I like this approach. If it's been tested properly, feel free to merge.

olivielpeau:
Yes, I've tested it on both the deb and rpm builds and it was working well.

Merging, I'll check the nightlies to make sure everything looks fine.

PR #58, opened 2016-01-19
Add datadogpy

### LINKED ISSUE #3: Adding dogapi?
It would be nice to have the ```dog``` cli script available which is provided by ~~https://github.com/DataDog/dogapi~~ https://github.com/DataDog/datadogpy - how about adding the python library to the omnibus distribution of dd-agent?

olivielpeau:
The `dog` and `dogwrap` scripts are shipped in `/opt/datadog-agent/bin/`.

To avoid conflicting with an existing install of datadogpy I haven't linked these scripts to `/usr/bin/` or to any other dir that is in the `PATH` by default, but we could do it if it makes sense.

Requires https://github.com/DataDog/omnibus-software/pull/32

Fixes #3

### DIFF
--- a/config/projects/datadog-agent.rb
+++ b/config/projects/datadog-agent.rb
@@ -153,6 +153,9 @@ dependency 'pyvmomi'
 dependency 'requests'
 dependency 'snakebite'
 
+# Additional software
+dependency 'datadogpy'
+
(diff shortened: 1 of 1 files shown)


remh:
:+1:

PR #61, opened 2016-02-01
[integrations] go-metro

truthbk:
Add `go-metro` dependencies to build system.

### DIFF
--- a/config/projects/datadog-agent.rb
+++ b/config/projects/datadog-agent.rb
@@ -161,6 +161,7 @@ dependency 'datadogpy'
 # be rebuilt (if put above, it would dirty the cache of the dependencies below
 # and trigger a useless rebuild of many packages)
 dependency 'datadog-gohai'
+dependency 'datadog-metro'
 
 # Datadog agent
 dependency 'datadog-agent'
--- a/omnibus_build.sh
+++ b/omnibus_build.sh
@@ -33,9 +33,10 @@ fi
 
 # Last but not least, let's make sure that we rebuild the agent everytime because
 # the extra package files are destroyed when the build container stops (we have
-# to tweak omnibus-git-cache directly for that). Same for gohai.
+# to tweak omnibus-git-cache directly for that). Same for gohai and go-metro.
 git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-agent`
 git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-gohai`
+git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-metro`
 
 # Install the gems we need, with stubs in bin/
 bundle update # Make sure to update to the latest version of omnibus-software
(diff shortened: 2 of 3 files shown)


package-scripts/datadog-agent/postinst
```diff
@@ -78,6 +78,9 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
   chown -R dd-agent:root /usr/bin/dd-agent
   chmod 755 /usr/bin/dd-agent
 
+  # Configuring capabilities for DD-TCP-RTT
+  setcap cap_net_raw=eip /opt/datadog-agent/bin/dd-tcp-rtt
```
truthbk: remove this - let customer run it manually.

config/projects/datadog-agent.rb
```diff
@@ -83,8 +83,10 @@
   # SysVInit service file
   if redhat?
     extra_package_file '/etc/rc.d/init.d/datadog-agent'
+    runtime_dependency 'libcap2'
   else
     extra_package_file '/etc/init.d/datadog-agent'
+    runtime_dependency 'libcap2-bin'
```
olivielpeau: You shouldn't declare `libcap` as a dependency of the `datadog-agent` package given that you embed the lib with `go-metro`. Tell me if I'm missing something though :)

omnibus_build.sh
```diff
@@ -33,9 +33,10 @@ fi
 
 # Last but not least, let's make sure that we rebuild the agent everytime because
 # the extra package files are destroyed when the build container stops (we have
-# to tweak omnibus-git-cache directly for that). Same for gohai.
+# to tweak omnibus-git-cache directly for that). Same for gohai and rtt.
 git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-agent`
 git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-gohai`
+git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-rtt`
```
olivielpeau: This should be updated to `datadog-metro` ;)

config/projects/datadog-agent.rb
truthbk: Hehe... yeah, it's not `libpcap` but `libcap` - this is the `capabilities` library that would be needed to set `CAP_NET_RAW` and allow traffic sniffing without being root. So this isn't at all PCAP related.

That said, we could make it a **soft** dependency and just instruct the customer to install it if they want to enable `go-metro`. Might be the safer course of action.

olivielpeau: Got it, thanks for the explanation. I agree that we shouldn't have `libcap` as a required dependency though.

package-scripts/datadog-agent/postinst
olivielpeau: the command is documented in the agent now, let's delete it here ;)

config/projects/datadog-agent.rb
olivielpeau: Let's remove this runtime dependency for both platforms now that you've documented them in go-metro's example yaml conf.

omnibus_build.sh
```diff
@@ -33,9 +33,10 @@ fi
 
 # Last but not least, let's make sure that we rebuild the agent everytime because
 # the extra package files are destroyed when the build container stops (we have
-# to tweak omnibus-git-cache directly for that). Same for gohai.
+# to tweak omnibus-git-cache directly for that). Same for gohai and rtt.
```
olivielpeau: nitpick: `rtt` -> `go-metro`

olivielpeau:
Just one nitpick, other than that let's :shipit:

PR #66, opened 2016-02-08
[python] Remove compiled python files from package

### LINKED ISSUE #64: 52 MB of Python byte code
@olivielpeau Please exclude Python byte code from the Debian packages.

`find /opt/datadog-agent "(" -name "*.pyc" -o -name "*.pyo" ")" -delete`

It takes up additional 52 MB.
Thank you.

olivielpeau:
We keep a list of the .pyc/.pyo files so that we can selectively
remove them in the pkg's prerm script.

Fixes #64

### DIFF
--- a/config/software/datadog-agent.rb
+++ b/config/software/datadog-agent.rb
@@ -53,8 +53,10 @@ build do
       command 'chmod 755 /etc/init.d/datadog-agent'
       touch '/usr/bin/dd-agent'
 
-      # Python-compile python's .py files so that the .pyc files have up-to-date timestamps with their .py counterparts
-      command "find #{install_dir}/embedded -name '*.pyc' | sed s/\\.pyc$/\\.py/ | #{install_dir}/embedded/bin/python -m compileall -q -i -"
+      # Remove the .pyc and .pyo files from the package and list them in a file
+      # so that the prerm script knows which compiled files to remove
+      command "echo '# DO NOT REMOVE/MODIFY - used by package removal tasks' > #{install_dir}/embedded/.py_compiled_files.txt"
+      command "find #{install_dir}/embedded '(' -name '*.pyc' -o -name '*.pyo' ')' -type f -delete -print >> #{install_dir}/embedded/.py_compiled_files.txt"
   end
 
   if osx?
--- a/package-scripts/datadog-agent/prerm
+++ b/package-scripts/datadog-agent/prerm
@@ -28,8 +28,14 @@ else
     exit 1;
 fi
 
-# Delete all.pyc files
+# Delete all.pyc files in the agent's dir
 find /opt/datadog-agent/agent -name '*.py[co]' -type f -delete || echo 'Unable to delete .pyc files'
 
-exit 0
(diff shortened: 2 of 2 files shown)


config/software/datadog-agent.rb
```diff
@@ -53,8 +53,10 @@
       command 'chmod 755 /etc/init.d/datadog-agent'
       touch '/usr/bin/dd-agent'
 
-      # Python-compile python's .py files so that the .pyc files have up-to-date timestamps with their .py counterparts
-      command "find #{install_dir}/embedded -name '*.pyc' | sed s/\\.pyc$/\\.py/ | #{install_dir}/embedded/bin/python -m compileall -q -i -"
+      # Remove the .pyc and .pyo files from the package, after having listed them to a file
+      # for the prerm script
+      command "find #{install_dir}/embedded -name '*.pyc' -o -name '*.pyo' > #{install_dir}/embedded/py_compiled_files.txt"
```
truthbk: open question: would a hidden file be better (given that users poke around)?

package-scripts/datadog-agent/prerm
```diff
@@ -28,8 +28,11 @@ else
...
 fi
 
-# Delete all.pyc files
+# Delete all.pyc files in the agent's dir
 find /opt/datadog-agent/agent -name '*.py[co]' -type f -delete || echo 'Unable to delete .pyc files'
 
-exit 0
+# Delete all the .pyc files in the embedded dir that are part of the agent's package
+# (i.e. not from packages installed manually by the user)
+cat /opt/datadog-agent/embedded/py_compiled_files.txt | xargs rm -f
```
truthbk: in the same vein as the comment above, what if some user has deleted the file?

config/software/datadog-agent.rb
olivielpeau: Using a hidden file would make sense yes

package-scripts/datadog-agent/prerm
olivielpeau: @truthbk We wouldn't delete anything in that case.

For a safer approach we could add a `grep /opt/datadog-agent/embedded/` filter on the contents of the file to make sure that we don't delete anything outside of the embedded dir.

But in general we're quite vulnerable to any change on the `py_compiled_files.txt` file.

truthbk: We should probably set permissions to `400 dd-agent:dd-agent`.

Also wondering if the removal command happens to fail (ie. file not found), would the uninstall operation be reported as a failure? I think it would, but not sure at all. Maybe worth wrapping it in a `File.exist?('/opt/datadog-agent/embedded/py_compiled_files.txt')`?

olivielpeau: Actually the permissions are `644 root:root`, we could restrict it more just to make it clear that the file is not meant to be modified.

And you're right currently the command would fail if the file isn't there. I can test that the file is there beforehand.

truthbk:
:shipit: :+1:

szepeviktor:
Thank you.

LeoCavaille:
:roller_coaster:

PR #69, opened 2016-02-26
Make go-metro Linux-specific

olivielpeau:
We also switch to a different supervisor.conf file because of the
related changes to supervisor.conf introduced by go-metro.

The `grep` on `supervisor.conf` is not necessary anymore (`user=dd-agent` has been removed on the related dd-agent PR).

Related to https://github.com/DataDog/dd-agent/pull/2297

### DIFF
--- a/config/projects/datadog-agent.rb
+++ b/config/projects/datadog-agent.rb
@@ -157,11 +157,13 @@ dependency 'snakebite'
 # Additional software
 dependency 'datadogpy'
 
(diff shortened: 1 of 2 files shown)


truthbk:
:shipit:

truthbk:
:+1: definitely a good move, having that `grep` was a bit of a hack. We could actually write up a generator for these files to feed off an XML or a JSON file to improve maintainability in the future.

PR #71, opened 2016-03-02
[prerm][python] Better handling of the pyc/pyo files removal on RPM

olivielpeau:
With RPM, if we remove the pyc/pyo files in the `prerm` script, when
the package is upgraded, we end up removing the files that are listed
in the `.py_compiled_files.txt` file of the _new_ version of the
package because the files of the new package get copied _before_ the
`prerm` script of the old package is run.

To avoid this, on RPM, we clean the pyc/pyo files in the prerm script
only when the package is removed. When the package is upgraded we do
the cleaning in the preinst script.

Behaviour unchanged for DEBs.

Also, updated docs.

### DIFF
--- a/package-scripts/datadog-agent/prerm
+++ b/package-scripts/datadog-agent/prerm
@@ -1,6 +1,16 @@
 #! /bin/sh
 LINUX_DISTRIBUTION=$(grep -Eo "(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)" /etc/issue)
 
+remove_py_compiled_files()
+{
+    # Delete all the .pyc files in the embedded dir that are part of the agent's package
+    if [ -f "/opt/datadog-agent/embedded/.py_compiled_files.txt" ]; then
+        # (commented lines are filtered out)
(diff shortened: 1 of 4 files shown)


package-scripts/datadog-agent/prerm
```diff
@@ -11,14 +21,19 @@ if [ -f "/etc/debian_version" ] || [ "$LINUX_DISTRIBUTION" == "Debian" ] || [ "$
     else
         /etc/init.d/datadog-agent stop || true
     fi
+
+    remove_py_compiled_files
```
truthbk: So this is this going to be run only for the uninstall on debian? Because we're already doing it in the `preinst`

olivielpeau: We do it in the `preinst` only for rpms (i.e. RHEL-like distros).

On debian the prerm of the old package is run on upgrades and uninstalls, and the script is always run first, so everything's pretty straightforward.

truthbk:
:shipit:

olivielpeau:
Thanks! Merging.

PR #76, opened 2016-05-05
[deb] Create dd-agent group in postinst script

olivielpeau:
No changes on the agent files' ownership/permissions for now (will be taken care of for 5.9.0).

Fixes https://github.com/DataDog/dd-agent/issues/1794

### DIFF
--- a/package-scripts/datadog-agent/postinst
+++ b/package-scripts/datadog-agent/postinst
@@ -41,15 +41,20 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
               update-rc.d datadog-agent defaults
               echo "Enabling service datadog-agent"
               update-rc.d datadog-agent enable >/dev/null 2>&1
-              # Only add dd-agent user if it doesn't already exist
+              # Only create dd-agent group and/or user if they don't already exist
+              getent group dd-agent >/dev/null || echo "Creating dd-agent group" && addgroup --system dd-agent --quiet
               set +e
               id -u dd-agent >/dev/null 2>&1
               USER_EXISTS=$?
               set -e
               if [ ! $USER_EXISTS -eq 0 ]; then
                   echo "Creating dd-agent user"
-                  adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
+                  adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --group --quiet
                   usermod -d /opt/datadog-agent dd-agent
+              elif id -nG dd-agent | grep --invert-match --word-regexp --quiet 'dd-agent'; then
+                  # User exists but is not part of the dd-agent group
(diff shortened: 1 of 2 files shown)


gmmeyer:
LGTM 👍

package-scripts/datadog-agent/postinst
```diff
@@ -41,15 +41,20 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
...
+              # Only create dd-agent group and/or user if they don't already exist
+              getent group dd-agent >/dev/null || echo "Creating dd-agent group" && addgroup --system dd-agent --quiet
               set +e
               id -u dd-agent >/dev/null 2>&1
               USER_EXISTS=$?
               set -e
               if [ ! $USER_EXISTS -eq 0 ]; then
                   echo "Creating dd-agent user"
-                  adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --quiet
+                  adduser --system dd-agent --disabled-login --shell /bin/sh --no-create-home --group --quiet
```
truthbk: you've probably taken a good look at the manpage, so I'm just saying this so we doublecheck - doesn't the `--group` flag take an argument after it? You are using the `--system` flag as well, so I'm not sure about this particular use-case.

package-scripts/datadog-agent/postinst
```diff
@@ -41,15 +41,20 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
...
                   usermod -d /opt/datadog-agent dd-agent
+              elif id -nG dd-agent | grep --invert-match --word-regexp --quiet 'dd-agent'; then
+                  # User exists but is not part of the dd-agent group
+                  echo "Adding dd-agent user to dd-agent group"
+                  usermod -g dd-agent dd-agent
```
truthbk: Wondering if the `dd-agent` user already exists and its main group is `X`, we might want to add 'X' to the list of supplementary groups with `-G 'X'`.

truthbk:
Added a couple of (very minor) comments/questions, I know this had been reviewed already, and already merged, but just adding for what it's worth.

package-scripts/datadog-agent/postinst
gmmeyer: It'll just add it to a group of the same name as the user (and create a group if necessary) if no other independent group name is given after the `--group` flag. I had to double check that after you said it, though.

package-scripts/datadog-agent/postinst
olivielpeau: good point, doing this would be less aggressive if users have already set up a group for the `dd-agent` user... We'd need to filter out the `nogroup` group though.

olivielpeau: I'll open a separate PR to improve this

PR #83, opened 2016-07-13
Uninstall fix

PranjalMishra:
Remove the /etc/logrotate.d/netsil-collectors file while purging the collectors.


remh:
Don't think it was meant to be open here but more likely there https://github.com/netsil/netsil-collectors-omnibus :)

PR #84, opened 2016-07-22
[rpm] Don't delete dd-agent user/group on uninstall

olivielpeau:


remh:
Looks good, let's make sure to test this properly (especially upgrades)

olivielpeau:
Tested installs, upgrades and downgrades, nothing to report!

PR #85, opened 2016-08-01
[debian] remove configcheck output when upgrading

degemer:
It's not useful for most users (especially when upgrading), and we ask
the user to run the command itself with the install script. Let's do the
same.


olivielpeau:
Makes sense and looks good, feel free to :shipit: !

PR #86, opened 2016-08-02
[linux] correctly silence configcheck

degemer:
These scripts use `/bin/sh` which doesn't support the `&>` syntax (bash
extension).

http://unix.stackexchange.com/a/80632


olivielpeau:
:o

LGTM

degemer:
🎉  @olivielpeau 🙇

PR #88, opened 2016-08-22
Integration SDK

gmmeyer:
This PR will update the Integration SDK.

Primarily it updates integration to require the datadog-agent as a dependency, and it updates the install_dir.

This will probably include more as the integration build pipeline gets more advanced, and I'll update this description as it includes more.

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -15,6 +15,8 @@ namespace :agent do
   task :'pull-integrations' do
     integration_branch = ENV['VERSION'] || 'master'
 
+    sh "ls -a #{ENV['INTEGRATIONS_REPO']}"
+    sh 'ls /'
     sh "git clone https://github.com/DataDog/#{ENV['INTEGRATIONS_REPO']}.git /#{ENV['INTEGRATIONS_REPO']} || true"
     sh "git --git-dir=/#{ENV['INTEGRATIONS_REPO']} fetch --all"
     sh "git --git-dir=/#{ENV['INTEGRATIONS_REPO']} checkout dd-check-#{ENV['INTEGRATION']}-#{integration_branch} ||
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -14,11 +14,9 @@ namespace :agent do
   desc 'Pull the integrations repo'
   task :'pull-integrations' do
     integration_branch = ENV['VERSION'] || 'master'
+
     sh "cd /#{ENV['INTEGRATIONS_REPO']} &&
-        git fetch --all &&
-        git checkout dd-check-#{ENV['INTEGRATION']}-#{integration_branch} &&
-        git checkout dd-check-#{ENV['INTEGRATION']}-#{integration_branch} &&
-        git reset --hard"
+        git fetch --all"
     sh "cd /#{ENV['INTEGRATIONS_REPO']} &&
         git checkout dd-check-#{ENV['INTEGRATION']}-#{integration_branch} ||
         git checkout #{integration_branch}"
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -15,10 +15,10 @@ namespace :agent do
   task :'pull-integrations' do
     integration_branch = ENV['VERSION'] || 'master'
 
-    sh "git -C /#{ENV['INTEGRATIONS_REPO']} fetch --all"
-    sh "git -C /#{ENV['INTEGRATIONS_REPO']} checkout dd-check-#{ENV['INTEGRATION']}-#{integration_branch} ||
-        git -C /#{ENV['INTEGRATIONS_REPO']} checkout #{integration_branch}"
-    sh "git -C /#{ENV['INTEGRATIONS_REPO']} reset --hard"
+    sh "git --git-dir=/#{ENV['INTEGRATIONS_REPO']} fetch --all"
+    sh "git --git-dir=/#{ENV['INTEGRATIONS_REPO']} checkout dd-check-#{ENV['INTEGRATION']}-#{integration_branch} ||
+        git --git-dir=/#{ENV['INTEGRATIONS_REPO']} checkout #{integration_branch}"
+    sh "git --git-dir=/#{ENV['INTEGRATIONS_REPO']} reset --hard"
   end
 
   desc 'Execute script'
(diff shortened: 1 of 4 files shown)


Rakefile.rb
```diff
@@ -12,9 +13,14 @@
 
   desc 'Pull the integrations repo'
   task :'pull-integrations' do
+    integration_branch = ENV['VERSION'] || 'master'
+
+    sh "cd /#{ENV['INTEGRATIONS_REPO']} &&
```
degemer: Maybe `Dir.chdir "/#{ENV['INTEGRATIONS_REPO']}"` instead ? But then we'd have to store the original repository.

Rakefile.rb
```diff
@@ -23,9 +29,12 @@
     sh "cd #{PROJECT_DIR} && bundle update"
     puts "building integration #{ENV['INTEGRATION']}"
 
+    manifest = JSON.parse(File.read("/#{ENV['INTEGRATIONS_REPO']}/#{ENV['INTEGRATION']}/manifest.json"))
+    integration_version = manifest['version'] || ENV['VERSION']
```
degemer: If there's a divergence, what does that mean ? What's right ? Or is it only for `ENV['VERSION'] == 'master'` ?

resources/datadog-integrations/project.rb.erb
```diff
@@ -43,8 +43,11 @@ end
 dependency 'preparation'  # creates required build directories
 dependency 'dd-check-<%= name %>-software'
 
+# Checks should require the datadog agent
+runtime_dependency 'datadog-agent'
```
degemer: Does it play nicely with both debs and rpms ? (I guess yes, just want to confirm)

Rakefile.rb
gmmeyer: It's for `ENV['VERSION'] == 'master'`

resources/datadog-integrations/project.rb.erb
gmmeyer: Yes, I believe so. I looked at how it's defined in the packager and tested it on both systems. We should exhaustively test this before we release. The tests do pass, so it does install fine.

Rakefile.rb
gmmeyer: We could also do 

```
git -C /#{ENV['INTEGRATIONS_REPO']} ...
```

gmmeyer: `git -C` might be the best solution. What do you think?

degemer: Yes, easiest solution and less dirty than a `cd`.

Rakefile.rb
degemer: Could we then do the json parsing only for `ENV['VERSION'] == 'master' ? (or `ENV['VERSION'] !~ /\d+\.\d+\.\d+/` ?

degemer: After an offline discussion, removing `ENV['VERSION']` is the best way to avoid confusion while still having a readable code. Every manifest should have a version, and every integration should have a manifest.

Rakefile.rb
gmmeyer: It's much cleaner with that

resources/datadog-integrations/project.rb.erb
```diff
@@ -43,8 +43,11 @@ end
...
+
 extends_packages('datadog-agent', 'datadog')
 
-extra_package_file '/etc/dd-agent/conf.d/<%= name %>.yaml.example'
+extra_package_file '/etc/dd-agent/conf.d/examples/<%= name %>.yaml.example'
```
degemer: What about default checks ? Will they only be shipped with the agent ? (or will the agent ship the `.yaml.default` and the integration package only a .example)

Rakefile.rb
```diff
@@ -12,20 +13,26 @@
 
   desc 'Pull the integrations repo'
   task :'pull-integrations' do
-    sh "cd /#{ENV['INTEGRATIONS_REPO']} &&
-        git fetch --all &&
-        git checkout dd-check-#{ENV['INTEGRATION']}-#{ENV['VERSION']} &&
-        git reset --hard"
+    integration_branch = ENV['VERSION'] || 'master'
+
+    sh "git -C /#{ENV['INTEGRATIONS_REPO']} fetch --all"
```
degemer: I actually found another way to do it in Ruby: http://stackoverflow.com/a/10148325.
It avoids the repetition of all the `-C` without having to run a `cd`.
Completely a nitpick though.

resources/datadog-integrations/project.rb.erb
irabinovitch: Seems like the package should include both, assuming its a package with default configs.

degemer: Checks with default config don't actually have an example, it's one or the other. But then the directory name is `examples`, not the best name if we want to put default configurations in there.

gmmeyer: @degemer @irabinovitch the checks that are shipped with the agent will continue to have all of their `conf.yaml.example` files reside in `/etc/dd-agent/conf.d`. The only ones that will have the config in `/etc/dd-agent/conf.d/examples` will be the ones installed via the SDK. Will any of these be the default ones?

Even if they are, we cannot overwrite the default config we already have, so we'll have to place it elsewhere. This leaves us with two options:

1. Do we drop them in examples, with the rest of them?

1. Do we add a new folder for them?

I think leaving them with the rest of the sdk ones makes the most sense. Doing it the other way would add another layer for the user to comprehend.

Rakefile.rb
gmmeyer: Do you think this is a better solution than `-C`? I am find handing this off to git rather than dealing with it in Ruby.

resources/datadog-integrations/project.rb.erb
degemer: You're right, defaults should be shipped with the agent, not with the checks themselves.
The only case I can see is if we publish a new package for a default check with a config change (because I think all packages will have their packages in the end, even if they are also shipped with the agent). But then I guess one user wanting to upgrade the check before the release of the agent will know the config changes to make. 
So I'd say that we don't have to ship defaults, only examples, and nothing to change.

irabinovitch: I dont think we can assume users who upgrade individual checks will only be expects who understand all the config options.  I'd push for all the examples to be in a single slot to avoid confusion and support pains. If that is conf.d/examples/ thats great if its all in conf.d thats also fine. 

Many of our challenges here are coming from choosing to ship all the integrations both in the agent itself, as well as in packages. As this results in both packages wreslting over ownership of a particular file.  If the agent package only installed the agent, but depending on the individual check packages we'd simplify this greatly.

gmmeyer: @irabinovitch tbh I agree with you however, the sdk installers cannot overwrite files (because the distros generally won't allow it).

PR #89, opened 2016-09-01
[apt] Only create dd-agent group if it doesn't already exist

suhasgaddam:
It looks like parentheses are needed for the appropriate order of operations.


olivielpeau:
Nice catch @suhasgaddam, thanks! You're right in `sh` the `&&` and `||` operators have the same precedence, so the parentheses are needed here.

FYI this change only affects the creation of the `dd-agent` group, I'll update the PR title accordingly. Also, the `addgroup` command returns `0` when the group already exists so this mistake in the script didn't have any user impact.

Merging, thanks again!

PR #90, opened 2016-09-06
Adds SUSE Enterprise support

gmmeyer:
This will allow us to build SUSE Enterprise packages.

### DIFF
--- a/config/software/datadog-agent.rb
+++ b/config/software/datadog-agent.rb
@@ -37,9 +37,9 @@ build do
   if linux?
     # Configuration files
     mkdir '/etc/dd-agent'
-    if ohai['platform_family'] == 'rhel'
+    if redhat?
       copy 'packaging/centos/datadog-agent.init', '/etc/rc.d/init.d/datadog-agent'
-    elsif ohai['platform_family'] == 'debian'
+    elsif debian?
       copy 'packaging/debian/datadog-agent.init', '/etc/init.d/datadog-agent'
       mkdir '/lib/systemd/system'
       copy 'packaging/debian/datadog-agent.service', '/lib/systemd/system/datadog-agent.service'
@@ -52,6 +52,7 @@ build do
       copy 'packaging/debian/start_agent.sh', '/opt/datadog-agent/bin/start_agent.sh'
       command 'chmod 755 /opt/datadog-agent/bin/start_agent.sh'
     end
+
     # Use a supervisor conf with go-metro on 64-bit platforms only
     if ohai['kernel']['machine'] == 'x86_64'
       copy 'packaging/supervisor.conf', '/etc/dd-agent/supervisor.conf'
--- a/config/software/datadog-agent.rb
+++ b/config/software/datadog-agent.rb
@@ -39,17 +39,20 @@ build do
     mkdir '/etc/dd-agent'
       if ohai['platform_family'] == 'rhel'
         copy 'packaging/centos/datadog-agent.init', '/etc/rc.d/init.d/datadog-agent'
-      elsif ohai['platform_family'] == 'debian' || suse?
+      elsif ohai['platform_family'] == 'debian'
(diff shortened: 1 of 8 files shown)


config/software/datadog-agent.rb
```diff
@@ -37,31 +37,38 @@
...
-      command "find #{install_dir}/embedded '(' -name '*.pyc' -o -name '*.pyo' ')' -type f -delete -print >> #{install_dir}/embedded/.py_compiled_files.txt"
+    if redhat?
+      copy 'packaging/centos/datadog-agent.init', '/etc/rc.d/init.d/datadog-agent'
+    elsif debian?
+      copy 'packaging/debian/datadog-agent.init', '/etc/init.d/datadog-agent'
+      mkdir '/lib/systemd/system'
+      copy 'packaging/debian/datadog-agent.service', '/lib/systemd/system/datadog-agent.service'
+      copy 'packaging/debian/start_agent.sh', '/opt/datadog-agent/bin/start_agent.sh'
+      command 'chmod 755 /opt/datadog-agent/bin/start_agent.sh'
+    elsif suse?
```
truthbk: most of this looks identical for `suse` and `debian` other than the destination [here](https://github.com/DataDog/dd-agent-omnibus/pull/90/files#diff-863803f6139877c9b5f8428ff98977b0R45), could we maybe condense this a little bit (I think it could help avoid discrepancies in the future)?

omnibus_build.sh
```diff
@@ -40,9 +40,9 @@ set +e
 # Last but not least, let's make sure that we rebuild the agent everytime because
 # the extra package files are destroyed when the build container stops (we have
 # to tweak omnibus-git-cache directly for that). Same for gohai and go-metro.
-git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-agent`
-git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-gohai`
-git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-metro`
+git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -d `git --git-dir=/var/cache/omnibus/cache/git_cache/opt/datadog-agent tag -l | grep datadog-agent` || true
```
truthbk: I think we don't need the `|| true` bit since we did `set +e` in line 38... Did you encounter issues here?

truthbk:
we probably want to squash these :)

config/software/datadog-agent.rb
gmmeyer: Yea I can do this.

omnibus_build.sh
gmmeyer: It's ugly and I know that it probably shouldn't, I couldn't figure out why it only does that on SUSE. If you have any insight into why, please let me know.

truthbk: @gmmeyer If you have the time, and can test easily, could you try removing the `set +e` and `set -e` we have on the top and bottom of this code block (it should be overkill now that we have the `|| true` part) and run with an empty/nonexisting cache?

gmmeyer: I can try it yes

gmmeyer: It works!

config/software/datadog-agent.rb
gmmeyer: @truthbk the logic is now conjoined

gmmeyer:
🎆

PR #96, opened 2016-10-19
[service discovery][grpc] generate client code module on the fly

truthbk:
This PR will generate the code for the GRPC client module on the fly and drop it in the relevant location in the embedded agent. We might need to set the `proto` file in a single source of truth location, because if I remember correctly, we still have it in two different places (JMXfetch and dd-agent repos). Should be an easy fix.


olivielpeau:
Closing this one too since we ended up not using grpc for Service Discovery in JMXFetch

PR #97, opened 2016-10-26
[postinst] Make `dd-agent` user and group own `/opt/datadog-agent`

olivielpeau:
### What this PR does

Makes `dd-agent` user and group own `/opt/datadog-agent` (which is the home directory of `dd-agent`)

### Motivation

Customer request: with this change, sysadmins can give no permissions at all to other users on the `/opt/datadog-agent` directory (required for PCI compliance), and the Agent can still work fine and not break the sysadmins' changes on upgrade.

### Testing

Tested the install of a local debian build: from scratch and as an upgrade ; both work fine.

### Additional notes

Also, on debian, set the home directory of `dd-agent` to
`/opt/datadog-agent` directly on the `adduser` call, for the sake of
brevity.

### DIFF
--- a/package-scripts/datadog-agent/postinst
+++ b/package-scripts/datadog-agent/postinst
@@ -49,8 +49,7 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
               set -e
               if [ ! $USER_EXISTS -eq 0 ]; then
                   echo "Creating dd-agent user"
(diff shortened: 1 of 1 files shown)


degemer:
Please add a big disclaimer in the CHANGELOG. People might not expect this to happen.

PR #99, opened 2016-11-08
Adds the new GPG key to the preinst script

gmmeyer:
Adds the new GPG key to the preinst script for Zypp and Yum based systems.

### DIFF
--- a/package-scripts/datadog-agent/preinst
+++ b/package-scripts/datadog-agent/preinst
@@ -46,7 +46,7 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
       useradd -r -M -g dd-agent -d $INSTALL_DIR -s /bin/sh \
         -c "Datadog Agent" dd-agent
     # Stop the old agent before installing
-    if [ -f "/etc/" ]; then
+    if [ -f "/etc/init.d/datadog-agent" ]; then
       /etc/init.d/datadog-agent stop || true
     fi
 
--- a/package-scripts/datadog-agent/preinst
+++ b/package-scripts/datadog-agent/preinst
@@ -54,6 +54,7 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
     if [ -f "/etc/yum.repos.d/datadog.repo" ]; then
       # only add the key if it's not there already
       if [ grep 'DATADOG_RPM_KEY_E09422B3.public' /etc/yum.repos.d/datadog.repo ]; then
+        echo "Prepare Datadog Agent keys rotation"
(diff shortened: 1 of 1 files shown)


package-scripts/datadog-agent/preinst
```diff
@@ -46,10 +46,27 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
       useradd -r -M -g dd-agent -d $INSTALL_DIR -s /bin/sh \
         -c "Datadog Agent" dd-agent
     # Stop the old agent before installing
-    if [ -f "/etc/init.d/datadog-agent" ]; then
```
degemer: Why change that ?

package-scripts/datadog-agent/preinst
```diff
@@ -46,10 +46,27 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
...
+    if [ -f "/etc/" ]; then
       /etc/init.d/datadog-agent stop || true
     fi
 
+    # Updates the yum repo to add the new GPG key for rhel based installations
```
degemer: We could maybe add the same disclaimer as Debian: https://github.com/DataDog/dd-agent-omnibus/blob/40a28d59096d1ba606494ad580ced856f6e2ad35/package-scripts/datadog-agent/preinst#L35-L38

package-scripts/datadog-agent/preinst
gmmeyer: I think I meant to copy something and cut instead, but not really sure how I did it, either way it's back.

gmmeyer:
@degemer I agree with you that this is problematic. A sha5 might work, it's probably the preferred solution here. What should we do for people who have a custom repo?

degemer:
Could we just append the new key at the end of the file, or is it not possible with the file format?
Side note: I thought `> file` failed when the file already existed.

gmmeyer:
woah thanks for following up! 

@degemer it does not, it overwrites

degemer:
Aaaand it seems that I always had noclobber option turned on.
So for now, I think it's a breaking change for people using their own repos (since it's going to overwrite their urls). But it's probably find since they're likely using Chef/Puppet/... to manage this file.
Summary:
- 95% of people with the default -> overwriting is fine
- 5% remaining -> maybe `grep datadoghq.com` ? If it's there, overwrite as usual, if not, they're using a custom repo and we shouldn't touch the file (hopefully managed by Chef/Puppet so already having our key).

gmmeyer:
this is dead I guess! closing

PR #100, opened 2016-11-09
[jmxfetch] build jmxfetch via omnibus

truthbk:
This PR adds the corresponding dependencies to enable building jmxfetch via omnibus-software as opposed to keeping the jar file in the `dd-agent` repo. It will also allow us to define what JMXfetch branch/tag we wish to include in the build, defaulting to `master` if none is specified.

https://github.com/DataDog/omnibus-software/pull/91

### DIFF
--- a/config/software/agent-deps.rb
+++ b/config/software/agent-deps.rb
@@ -31,10 +31,9 @@ end
 dependency 'boto'
 dependency 'docker-py'
 
-if ENV['JMX_VERSION'] && !ENV['JMX_VERSION'].empty?
-  dependency 'jmxfetch'
-  dependency 'jmxterm'
-end
+dependency 'jmxfetch'
+dependency 'jmxterm'
+
 dependency 'ntplib'
 dependency 'protobuf-py'
 dependency 'psutil'
--- a/config/software/agent-deps.rb
+++ b/config/software/agent-deps.rb
@@ -30,6 +30,11 @@ end
 # Agent dependencies
 dependency 'boto'
 dependency 'docker-py'
+
+if ENV['JMX_VERSION'] && !ENV['JMX_VERSION'].empty?
+  dependency 'jmxfetch'
+  dependency 'jmxterm'
+end
 dependency 'ntplib'
 dependency 'protobuf-py'
 dependency 'psutil'
--- a/omnibus_build.sh
+++ b/omnibus_build.sh
@@ -9,7 +9,6 @@
 PROJECT_DIR=dd-agent-omnibus
 PROJECT_NAME=datadog-agent
 LOG_LEVEL=${LOG_LEVEL:-"info"}
-JMX_VERSION=${JMX_BRANCH:-"current"}
 export AGENT_BRANCH=${AGENT_BRANCH:-"master"}
 export OMNIBUS_BRANCH=${OMNIBUS_BRANCH:-"master"}
 export OMNIBUS_SOFTWARE_BRANCH=${OMNIBUS_SOFTWARE_BRANCH:-"master"}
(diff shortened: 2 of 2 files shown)


omnibus_build.sh
```diff
@@ -9,11 +9,22 @@
...
 export OMNIBUS_BRANCH=${OMNIBUS_BRANCH:-"master"}
 export OMNIBUS_SOFTWARE_BRANCH=${OMNIBUS_SOFTWARE_BRANCH:-"master"}
 export OMNIBUS_RUBY_BRANCH=${OMNIBUS_RUBY_BRANCH:-"datadog-5.5.0"}
 export INTEGRATION_CORE_BRANCH=${INTEGRATION_CORE_BRANCH:-"master"}
 
+REMOTE_AGENT_REPO_RAW="https://raw.githubusercontent.com/DataDog/dd-agent"
+LOCAL_DD_AGENT="/dd-agent-repo"
+
+unset_jmx() {
+    echo "JMX provided by dd-agent in selected branch, JMX environment variables will be ignored."
```
olivielpeau: Should we still support this case once the PR is merged? Shouldn't the jar always be pulled from the s3 bucket?
Removing support for this case would largely simplify the code below IMO, we could just fail the build if we can't find the version of jmxfetch in the dd-agent file.
What do you think?

truthbk: I'm not sure what to do with this. On one hand you're absolutely right. On the other hand, we should be able to build to build any dd-agent branch on cue. We'd have to upload all jmxfetch releases to the bucket, which isn't much of a problem, but the jarfiles are already part of the repo,... not sure.

truthbk: So, since indeed the `JMX_VERSION` will just be ignored with other pinned versions, we don't need to unset anything. We can cleanup the script.

omnibus_build.sh
```diff
@@ -9,11 +9,16 @@
 PROJECT_DIR=dd-agent-omnibus
 PROJECT_NAME=datadog-agent
 LOG_LEVEL=${LOG_LEVEL:-"info"}
+JMX_VERSION=${JMX_BRANCH:-"current"}
```
olivielpeau: as discussed let's remove this altogether: the jmxfetch version should be the one that's defined in the agent's `config.py` on the branch that we're building.

config/software/agent-deps.rb
```diff
@@ -30,6 +30,11 @@
 # Agent dependencies
 dependency 'boto'
 dependency 'docker-py'
+
+if ENV['JMX_VERSION'] && !ENV['JMX_VERSION'].empty?
```
olivielpeau: I think we can remove this (for simplicity): when we merge this PR to master, `dd-agent` should be ready for the new jmxfetch build, and if we want to build agents based on bugfix branches we can use the bugfix branches of `dd-agent-omnibus`.

Let me know if you disagree :)

omnibus_build.sh
```diff
@@ -31,11 +35,23 @@ git fetch --all
...
+# NOTE: JMX_VERSION plays no role <5.14.x (jmx bundled with agent)
+# If "current" JMX version, set it from the corresponding agent branch config.py
+if [ -n "$LOCAL_AGENT_REPO" ]; then
+  cd $LOCAL_DD_AGENT
+  JMX_VERSION=$(git show $AGENT_BRANCH:config.py | grep 'JMX_VERSION' | cut -f2 -d'=' | tr -d ' "')
+  cd -
+else
+  JMX_VERSION=$(curl -v $REMOTE_AGENT_REPO_RAW/$AGENT_BRANCH/config.py 2>/dev/null | grep 'JMX_VERSION' | cut -f2 -d'=' | tr -d ' "')
+fi
+export JMX_VERSION=$JMX_VERSION
```
olivielpeau: nit: just export JMX_VERSION works no?

omnibus_build.sh
```diff
@@ -9,11 +9,15 @@
 PROJECT_DIR=dd-agent-omnibus
 PROJECT_NAME=datadog-agent
 LOG_LEVEL=${LOG_LEVEL:-"info"}
+export AGENT_BRANCH=${AGENT_BRANCH:-"master"}
```
olivielpeau: qq: why did you add this export? (just interested in knowing why :) )

truthbk: I added it, so the user could specify a branch for the local repo if he so wished (ie. a local dev branch).

PR #104, opened 2016-12-12
Fixes issues with logfile and tmpfiles

gmmeyer:
This will add the logfiles under /opt/datadog-agent/log (or equivalent) and symlink the directory to the current system log directory (`ln -s /opt/datadog-agent/log /var/log/datadog`).

It will also create the tmp_dir for osx.


gmmeyer:
No longer of any immediate need, closing.

PR #106, opened 2017-01-09
Pulls in checks from the integration core repo

gmmeyer:
This changes the logic of the build to ensure that the integrations are built from the integration sdk repo and not from the dd-agent repo.

### DIFF
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -68,20 +68,6 @@ build do
       copy "#{integrations_dir}/#{check}/conf.yaml.default", "#{conf_directory}/#{check}.yaml.default"
     end
 
-    # There are some cases where there are multiple config files for a check,
-    # It's mostly for ones where we have both an agent check and a jmx check,
-    # And by some, I mean only ActiveMQ
-    if Dir.exists? "#{integrations_dir}/#{check}"
-      Dir.foreach "#{integrations_dir}/#{check}/conf" do |config_file|
-        config_file.slice! "#{integrations_dir}/#{check}/conf/"
-        if config_file == 'auto_conf.yaml'
-          copy "#{integrations_dir}/#{check}/conf/#{config_file}", "#{conf_directory}/auto_conf/#{check}.yaml"
-        else
-          copy "#{integrations_dir}/#{check}/conf/#{config_file}", "#{conf_directory}/#{config_file}"
-        end
-      end
-    end
-
     # We don't have auto_conf on windows yet
     unless windows?
       if File.exists? "#{integrations_dir}/#{check}/auto_conf.yaml"
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -100,5 +100,5 @@ build do
   # Close the checks requirements file
   all_reqs_file.close
 
-  pip "install --target=#{install_dir}/lib -c #{install_dir}/agent/requirements.txt -r #{install_dir}/agent/check_requirements.txt", :env => env
+  pip "install -c #{install_dir}/agent/requirements.txt -r #{install_dir}/agent/check_requirements.txt", :env => env
 end
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -7,10 +7,6 @@ dependency 'datadog-agent'
 
 relative_path 'integrations-core'
 
-env = {
-  "PATH" => "#{install_dir}/embedded/bin:#{ENV['PATH']}",
-}
-
 # The only integrations that will be packaged with the agent
 # are the ones that are officiallly supported.
 local_integrations_core_repo = ENV['LOCAL_INTEGRATIONS_CORE_REPO']
@@ -29,7 +25,6 @@ else
 end
 
 build do
-  mkdir  "#{install_dir}/agent/"
   # Agent code
   mkdir "#{install_dir}/agent/checks.d"
 
@@ -86,5 +81,5 @@ build do
   # Close the checks requirements file
   all_reqs_file.close
 
-  pip "install -c #{install_dir}/agent/requirements.txt -r #{install_dir}/agent/check_requirements.txt", :env => env
+  pip "install --install-option=\"--install-scripts=#{windows_safe_path(install_dir)}/bin\" -c #{install_dir}/agent/requirements.txt -r #{install_dir}/agent/check_requirements.txt"
 end
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -3,6 +3,7 @@ require './lib/ostools.rb'
 name 'datadog-agent-integrations'
 
 dependency 'pip'
+dependency 'datadog-agent'
 
 relative_path 'integrations-core'
 
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -34,29 +34,28 @@ build do
 
   # Grab all the checks
   checks = Dir.glob("#{integrations_dir}/*/")
+  # Only use the parts of the filename we need
+  check.slice! "#{integrations_dir}/"
+  check.slice! "/"
 
   # Open the concatenated checks requirements file
   # We're going to store it with the agent install
   all_reqs_file = File.open("#{install_dir}/agent/check_requirements.txt", 'w')
 
+  # The conf directory is different on every system
+  if linux?
+    conf_directory = "/etc/dd-agent/conf.d"
+  elsif osx?
+    conf_directory = "#{install_dir}/etc"
+  elsif windows?
+    conf_directory = "../../extra_package_files/EXAMPLECONFSLOCATION"
+  end
+
   # loop through them
   checks.each do |check|
-    # Only use the parts of the filename we need
-    check.slice! "#{integrations_dir}/"
-    check.slice! "/"
-
     # Copy the checks over
     if File.exists? "#{integrations_dir}/#{check}/check.py"
       copy "#{integrations_dir}/#{check}/check.py", "#{install_dir}/agent/checks.d/#{check}.py"
-    end
-
-    # The conf directory is different on every system
-    if linux?
-      conf_directory = "/etc/dd-agent/conf.d"
-    elsif osx?
-      conf_directory = "#{install_dir}/etc"
-    elsif windows?
-      conf_directory = "../../extra_package_files/EXAMPLECONFSLOCATION"
     end
 
     # Copy the check config to the conf directories
@@ -72,7 +71,7 @@ build do
     # It's mostly for ones where we have both an agent check and a jmx check,
     # And by some, I mean only ActiveMQ
     if Dir.exists? "#{integrations_dir}/#{check}"
-      Dir.foreach "#{integrations_dir}/#{check}" do |config_file|
+      Dir.foreach "#{integrations_dir}/#{check}/conf" do |config_file|
         config_file.slice! "#{integrations_dir}/#{check}/conf/"
         if config_file == 'auto_conf.yaml'
           copy "#{integrations_dir}/#{check}/conf/#{config_file}", "#{conf_directory}/auto_conf/#{check}.yaml"
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -30,9 +30,6 @@ build do
 
   # Grab all the checks
   checks = Dir.glob("#{integrations_dir}/*/")
-  # Only use the parts of the filename we need
-  check.slice! "#{integrations_dir}/"
-  check.slice! "/"
 
   # Open the concatenated checks requirements file
   # We're going to store it with the agent install
@@ -49,6 +46,10 @@ build do
 
   # loop through them
   checks.each do |check|
+    # Only use the parts of the filename we need
+    check.slice! "#{integrations_dir}/"
+    check.slice! "/"
+
     # Copy the checks over
     if File.exists? "#{integrations_dir}/#{check}/check.py"
       copy "#{integrations_dir}/#{check}/check.py", "#{install_dir}/agent/checks.d/#{check}.py"
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -1,0 +1,100 @@
+require './lib/ostools.rb'
+
+name 'datadog-agent-integrations'
+
+dependency 'pip'
+
+relative_path 'integrations-core'
+
+env = {
+  "PATH" => "#{install_dir}/embedded/bin:#{ENV['PATH']}",
+}
+
+# The only integrations that will be packaged with the agent
+# are the ones that are officiallly supported.
+local_integrations_core_repo = ENV['LOCAL_INTEGRATIONS_CORE_REPO']
+if local_integrations_core_repo.nil? || local_integrations_core_repo.empty?
+  source git: 'https://github.com/DataDog/integrations-core.git'
+else
+  # For local development
+  source path: ENV['LOCAL_INTEGRATIONS_CORE_REPO']
+end
+
+integrations_core_branch = ENV['INTEGRATION_CORE_BRANCH']
+if integrations_core_branch.nil? || integrations_core_branch.empty?
+  default_version 'master'
+else
+  default_version integrations_core_branch
+end
+
+build do
+  if ohai['platform'] == "windows"
+    integrations_dir = project_dir
+  else
+    integrations_dir = project_dir
+  end
+
+  mkdir  "#{install_dir}/agent/"
+  # Agent code
+  mkdir "#{install_dir}/agent/checks.d"
+
+  # Grab all the checks
+  checks = Dir.glob("#{integrations_dir}/*/")
+  # logging.info('datadog-agent-integrations') { checks }
+  # logging.info('datadog-agent-integrations') { Dir.glob("*") }
+
+  # loop through them
+  checks.each do |check|
+    # Only use the parts of the filename we need
+    check.slice! "#{integrations_dir}/"
+    check.slice! "/"
+
+    # Copy the checks over
+    if File.exists? "#{integrations_dir}/#{check}/check.py"
+      copy "#{integrations_dir}/#{check}/check.py", "#{install_dir}/agent/checks.d/#{check}.py"
+    end
+
+    # The conf directory is different on every system
+    if linux?
+      conf_directory = "/etc/dd-agent/conf.d"
+    elsif osx?
+      conf_directory = "#{install_dir}/etc"
+    elsif windows?
+      conf_directory = "../../extra_package_files/EXAMPLECONFSLOCATION"
+    end
+
+    # Copy the check config to the conf directories
+    if File.exists? "#{integrations_dir}/#{check}/conf.yaml.example"
+      copy "#{integrations_dir}/#{check}/conf.yaml.example", "#{conf_directory}/#{check}.yaml.example"
+    end
+    # Copy the default config, if it exists
+    if File.exists? "#{integrations_dir}/#{check}/conf.yaml.default"
+      copy "#{integrations_dir}/#{check}/conf.yaml.default", "#{conf_directory}/#{check}.yaml.default"
+    end
+
+    # There are some cases where there are multiple config files for a check,
+    # It's mostly for ones where we have both an agent check and a jmx check,
+    # And by some, I mean only ActiveMQ
+    if Dir.exists? "#{integrations_dir}/#{check}"
+      Dir.foreach "#{integrations_dir}/#{check}" do |config_file|
+        config_file.slice! "#{integrations_dir}/#{check}/conf/"
+        if config_file == 'auto_conf.yaml'
+          copy "#{integrations_dir}/#{check}/conf/#{config_file}", "#{conf_directory}/auto_conf/#{check}.yaml"
+        else
+          copy "#{integrations_dir}/#{check}/conf/#{config_file}", "#{conf_directory}/#{config_file}"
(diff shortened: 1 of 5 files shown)


config/software/datadog-agent.rb
```diff
@@ -10,6 +10,12 @@
   source path: ENV['LOCAL_AGENT_REPO']
 end
 
+if ohai['platform'] == "windows"
+  integrations_dir="../integrations-core"
```
degemer: Nit: can you add a space before and after the `=` sign? @olivielpeau should we run rubocop (with the minimum of rules)

config/software/datadog-agent.rb
```diff
@@ -25,7 +31,18 @@
   mkdir  "#{install_dir}/agent/"
 
   # Agent code
+  mkdir "#{install_dir}/agent/checks.d"
   copy 'checks.d', "#{install_dir}/agent/"
+  checks = Dir["#{integrations_dir}/*"]
+  if ENV["USE_INTEGRATION_SDK"]
```
degemer: What do you think of checking the existence of `integrations_dir` instead? I was thinking that this could also protect us from errors if the directory doesn't exist, but since you use `Dir["#{integrations_dir}/*"]` we should be 🆗 .

config/software/datadog-agent.rb
```diff
@@ -64,7 +81,21 @@
       copy 'packaging/supervisor_32.conf', '/etc/dd-agent/supervisor.conf'
     end
     copy 'datadog.conf.example', '/etc/dd-agent/datadog.conf.example'
+    mkdir "/etc/dd-agent/conf.d/auto_conf"
```
degemer: `auto_conf` is used there, whereas a few lines after it's `autoconf`. Which one is the good one?

config/software/datadog-agent.rb
```diff
@@ -64,7 +81,21 @@
...
     copy 'conf.d', '/etc/dd-agent/'
+    checks = Dir["#{integrations_dir}/*"]
+    if ENV["USE_INTEGRATION_SDK"]
+      checks.each do |check|
+        check.slice! "#{integrations_dir}/"
+        if File.exists? "#{integrations_dir}/#{check}/autoconf.yaml"
+          copy "#{integrations_dir}/#{check}/autoconf.yaml", "/etc/dd-agent/conf.d/autoconf/#{check}.yaml"
```
degemer: Are we getting rid of the `{check}.yaml.default` ? (or is it something else)

config/software/datadog-agent.rb
```diff
@@ -137,7 +168,20 @@
     mkdir "#{install_dir}/etc"
     copy "packaging/osx/supervisor.conf", "#{install_dir}/etc/supervisor.conf"
     copy 'datadog.conf.example', "#{install_dir}/etc/datadog.conf.example"
+    mkdir "/etc/dd-agent/conf.d/auto_conf"
```
degemer: Same, `auto_conf` vs `autoconf`.

config/software/datadog-agent.rb
```diff
@@ -25,7 +31,18 @@
   mkdir  "#{install_dir}/agent/"
 
   # Agent code
+  mkdir "#{install_dir}/agent/checks.d"
   copy 'checks.d', "#{install_dir}/agent/"
+  checks = Dir["#{integrations_dir}/*"]
```
degemer: This will pick up files as well, not only directories. `Dir.glob("#{integrations_dir}/*/"` should give you what you expect. (I prefer `Dir.glob` over `Dir[]` because it's easier to understand when you don't know Ruby)

config/software/datadog-agent.rb
gmmeyer: It exists now

config/software/datadog-agent.rb
gmmeyer: This is a bad typo that I have rectified

config/software/datadog-agent.rb
gmmeyer: I am not sure, something to discuss, good point

config/software/datadog-agent.rb
gmmeyer: Yes this is a good point, I'll make that change

config/software/datadog-agent.rb
gmmeyer: I can do whatever you want, and yea that's definitely something that's a good cosmetic issue

gmmeyer:
On a suggestion from @remh, I moved the logic out to it's own software definition. It's much cleaner and easier to read now

omnibus_build.sh
```diff
@@ -35,6 +37,12 @@ if [ -n "$RPM_SIGNING_PASSPHRASE" ]; then
   gpg --import /keys/RPM-SIGNING-KEY.private
 fi
 
+git --git-dir=/integrations-core fetch --all || true
+git --git-dir=/integrations-core checkout $INTEGRATION_CORE_BRANCH || true
```
olivielpeau: I don't understand why we're fetching checking out the `integrations-core` repo here, we should probably remove these lines no?

omnibus_build.sh
```diff
@@ -35,6 +37,12 @@ if [ -n "$RPM_SIGNING_PASSPHRASE" ]; then
...
+
+echo "integrations"
+ls /integrations-core
```
olivielpeau: let's remove this from here, if we want to log something about the integrations we're building I think we should do it from the `build` step of `datadog-agent-integrations` with `Omnibus.logger.debug("")` for instance

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,67 @@
+require './lib/ostools.rb'
+
+name 'datadog-agent-integrations'
+
```
olivielpeau: let's declare the dependency on `datadog-agent`

olivielpeau:
One other thing: we have to clean up the pip dependencies defined in https://github.com/DataDog/dd-agent-omnibus/blob/5.10.1/config/projects/datadog-agent.rb#L131-L168 since most of them are going to be defined in `integrations-core` and should be installed in the `build` step of `datadog-agent-integrations`.

Some of the pip deps currently defined in their own software defs are more than just a simple `pip install`, let's be extra-careful on these ones.

We need to confirm that we're shipping the same versions of the same deps with `integrations-core` compared to what we currently ship with `dd-agent`, open to discussion on how we could check that.

config/software/datadog-agent-integrations.rb
gmmeyer: This is a part of building the datadog-agent. In the definition of the independent checks we do declare it: https://github.com/DataDog/dd-agent-omnibus/blob/master/resources/datadog-integrations/project.rb.erb#L47

gmmeyer:
We definitely need to confirm that. We currently have a task that does this to some extent:

https://github.com/DataDog/datadog-sdk-testing/blob/master/lib/tasks/ci/hooks/pre-commit.py

However, it is not perfect and it _does not_ check them against the requirements defined here or in omnibus-software. And, further, it does not ensure that there are no conflicts between integrations core and integrations-extra

config/software/datadog-agent.rb
gmmeyer: c

config/software/datadog-agent.rb
gmmeyer: it's auto_conf, I fixed all these I think now

config/software/datadog-agent-integrations.rb
degemer: I think this only declares a dependency for the package itself (i.e. it needs `datadog-agent`), whereas I think what Olivier wants is a dependency in the Omnibus sense (so that we're sure that `datadog-agent-integrations` is built after `datadog-agent`).

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,100 @@
...
+}
+
+# The only integrations that will be packaged with the agent
+# are the ones that are officiallly supported.
+local_integrations_core_repo = ENV['LOCAL_INTEGRATIONS_CORE_REPO']
+if local_integrations_core_repo.nil? || local_integrations_core_repo.empty?
+  source git: 'https://github.com/DataDog/integrations-core.git'
+else
+  # For local development
+  source path: ENV['LOCAL_INTEGRATIONS_CORE_REPO']
```
degemer: 🍰

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,100 @@
...
+
+integrations_core_branch = ENV['INTEGRATION_CORE_BRANCH']
+if integrations_core_branch.nil? || integrations_core_branch.empty?
+  default_version 'master'
+else
+  default_version integrations_core_branch
+end
+
+build do
+  if ohai['platform'] == "windows"
```
degemer: Hum, doesn't seem needed right now?

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  mkdir  "#{install_dir}/agent/"
+  # Agent code
+  mkdir "#{install_dir}/agent/checks.d"
+
+  # Grab all the checks
+  checks = Dir.glob("#{integrations_dir}/*/")
+  # logging.info('datadog-agent-integrations') { checks }
+  # logging.info('datadog-agent-integrations') { Dir.glob("*") }
```
degemer: Debug logs, let's remove them?

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,100 @@
...
+    # Only use the parts of the filename we need
+    check.slice! "#{integrations_dir}/"
+    check.slice! "/"
+
+    # Copy the checks over
+    if File.exists? "#{integrations_dir}/#{check}/check.py"
+      copy "#{integrations_dir}/#{check}/check.py", "#{install_dir}/agent/checks.d/#{check}.py"
+    end
+
+    # The conf directory is different on every system
```
degemer: We could do it outside of the checks.each loop

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,100 @@
...
+  # logging.info('datadog-agent-integrations') { Dir.glob("*") }
+
+  # loop through them
+  checks.each do |check|
+    # Only use the parts of the filename we need
+    check.slice! "#{integrations_dir}/"
+    check.slice! "/"
+
+    # Copy the checks over
+    if File.exists? "#{integrations_dir}/#{check}/check.py"
```
degemer: Shouldn't we do the opposite? I.e. if `check.py` doesn't exist, log an error and skip (or fail)?

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,100 @@
...
+    # Copy the default config, if it exists
+    if File.exists? "#{integrations_dir}/#{check}/conf.yaml.default"
+      copy "#{integrations_dir}/#{check}/conf.yaml.default", "#{conf_directory}/#{check}.yaml.default"
+    end
+
+    # There are some cases where there are multiple config files for a check,
+    # It's mostly for ones where we have both an agent check and a jmx check,
+    # And by some, I mean only ActiveMQ
+    if Dir.exists? "#{integrations_dir}/#{check}"
+      Dir.foreach "#{integrations_dir}/#{check}" do |config_file|
```
degemer: Missing a `/conf` at the end?

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,100 @@
...
+        config_file.slice! "#{integrations_dir}/#{check}/conf/"
+        if config_file == 'auto_conf.yaml'
+          copy "#{integrations_dir}/#{check}/conf/#{config_file}", "#{conf_directory}/auto_conf/#{check}.yaml"
```
degemer: Couldn't we always use the `conf` directory then? Or never use it? It seems weird to add it for one integration.

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    # We don't have auto_conf on windows yet
+    unless windows?
+      if File.exists? "#{integrations_dir}/#{check}/auto_conf.yaml"
+        copy "#{integrations_dir}/#{check}/autoconf.yaml", "#{conf_directory}/auto_conf/#{check}.yaml"
+      end
+    end
+
+    if File.exists? "#{integrations_dir}/#{check}/requirements.txt"
+      pip "install -I --target=#{install_dir}/lib -r #{integrations_dir}/#{check}/requirements.txt", :env => env
```
degemer: Shouldn't the target contain the check name? (for each check to have its own libs)

degemer: Shouldn't we use `--prefix`? (Open question) https://pip.pypa.io/en/latest/reference/pip_install/#cmdoption--prefix

degemer: This is only run when we want to ship the checks within the agent? If yes, then we shouldn't have to specify the target. (I think)

resources/datadog-integrations/software.rb.erb
```diff
@@ -13,9 +13,46 @@ build do
...
+  mkdir "#{conf_directory}/examples/"
+
+  # Move config files to examples folder.
+  if File.exists? "#{integrations_dir}/#{check}/conf.yaml.example"
+    copy "#{integration_source_folder}/conf.yaml.example", "/etc/dd-agent/conf.d/examples/<%= name %>.yaml.example"
+  end
+
+  # If it is a default, we need to move it as a default
+  if File.exists? "#{integrations_dir}/#{check}/conf.yaml.default"
+    copy "#{integration_source_folder}/conf.yaml.default", "/etc/dd-agent/conf.d/examples/<%= name %>.yaml.default"
```
degemer: This is a real issue. If we change a default and the customer wants to get the last changes, he'll have to change the file manually. I can't think of a good solution right now though...

config/software/datadog-agent-integrations.rb
olivielpeau: let's remove the `-I` option here, I don't think we should ever use it  (applies to the software def template of the individual integration packages as well)

config/software/datadog-agent-integrations.rb
gmmeyer: Oh, well how do I do this @degemer ?

resources/datadog-integrations/software.rb.erb
gmmeyer: Yea, I don't know the answer, either.

config/software/datadog-agent-integrations.rb
gmmeyer: it's gone you're right

config/software/datadog-agent-integrations.rb
gmmeyer: Already done!

config/software/datadog-agent-integrations.rb
gmmeyer: JMX checks don't have a check.py

config/software/datadog-agent-integrations.rb
gmmeyer: You're right

config/software/datadog-agent-integrations.rb
gmmeyer: Yes I am, thank you

config/software/datadog-agent-integrations.rb
gmmeyer: I'd prefer to not use it if we don't have to **shrug**

config/software/datadog-agent-integrations.rb
gmmeyer: Done

config/software/datadog-agent-integrations.rb
degemer: `dependency 'datadog-agent'`, before/after `dependency 'pip'` should do the trick.

config/software/datadog-agent-integrations.rb
degemer: Aaaah, forgot about them, sorry!

gmmeyer:
Also @degemer I agree that the activemq thing is not great. We might just deprecate one of the yaml files.

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,104 @@
...
+
+name 'datadog-agent-integrations'
+
+dependency 'pip'
+dependency 'datadog-agent'
+
+relative_path 'integrations-core'
+
+env = {
+  "PATH" => "#{install_dir}/embedded/bin:#{ENV['PATH']}",
```
olivielpeau: why set a custom `PATH`here?

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,104 @@
...
+      reqs.each_line do |line|
+        all_reqs_file << line
+      end
+    end
+  end
+
+  # Close the checks requirements file
+  all_reqs_file.close
+
+  pip "install -c #{install_dir}/agent/requirements.txt -r #{install_dir}/agent/check_requirements.txt", :env => env
```
olivielpeau: if we want to be as close as possible to the current way we install pip deps we should probably use the `--install-option=\"--install-scripts=#{windows_safe_path(install_dir)}/bin\"` option

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,104 @@
...
+
+integrations_core_branch = ENV['INTEGRATION_CORE_BRANCH']
+if integrations_core_branch.nil? || integrations_core_branch.empty?
+  default_version 'master'
+else
+  default_version integrations_core_branch
+end
+
+build do
+  mkdir  "#{install_dir}/agent/"
```
olivielpeau: hmm this shouldn't be necessary

resources/datadog-integrations/software.rb.erb
```diff
@@ -13,13 +13,50 @@ build do
...
+
+  # We need to move all the important stuff over
+  # The manifest and the readme are important to move over
   for file in ["check.py", "manifest.json", "requirements.txt", "README.md"]
     copy "#{integration_source_folder}/#{file}", "#{integration_agent_folder}/#{file}"
   end
 
   # install requirements
-  command "#{install_dir}/embedded/bin/pip install -I --target=#{integration_agent_folder}/lib -r #{integration_source_folder}/requirements.txt"
+  command "#{install_dir}/embedded/bin/pip install --target=#{integration_agent_folder}/lib -c -c #{install_dir}/agent/requirements.txt -r #{integration_source_folder}/requirements.txt"
```
olivielpeau: let's use the `pip` DSL method :)

config/software/datadog-agent-integrations.rb
gmmeyer: no, it's not

config/software/datadog-agent-integrations.rb
gmmeyer: Good idea

resources/datadog-integrations/software.rb.erb
degemer: Let's say that we don't ship default files then? And only ship examples (converting them to .example if needed). It's the easiest solution, and really a edge case.

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,85 @@
...
+end
+
+build do
+  # Agent code
+  mkdir "#{install_dir}/agent/checks.d"
+
+  # Grab all the checks
+  checks = Dir.glob("#{integrations_dir}/*/")
+  # Only use the parts of the filename we need
+  check.slice! "#{integrations_dir}/"
```
degemer: Not the right place, needs to be after `checks.each`? Or can you apply `slice` to each `checks` globally?

config/software/datadog-agent-integrations.rb
```diff
@@ -0,0 +1,85 @@
...
+    unless windows?
+      if File.exists? "#{integrations_dir}/#{check}/auto_conf.yaml"
+        copy "#{integrations_dir}/#{check}/autoconf.yaml", "#{conf_directory}/auto_conf/#{check}.yaml"
+      end
+    end
+
+    if File.exists? "#{integrations_dir}/#{check}/requirements.txt"
+      reqs = File.open(file, 'r').read
+      reqs.each_line do |line|
+        all_reqs_file << line
```
degemer: Does pip play well when multiple integrations declare the same reqs?

config/software/datadog-agent.rb
```diff
@@ -25,11 +25,15 @@
...
 
   # Agent code
+  mkdir "#{install_dir}/agent/checks.d"
   copy 'checks.d', "#{install_dir}/agent/"
+
   copy 'checks', "#{install_dir}/agent/"
   copy 'dogstream', "#{install_dir}/agent/"
   copy 'utils', "#{install_dir}/agent/"
 
+  copy 'requirements.txt', "#{install_dir}/agent/"
```
degemer: Do we need it?

config/software/datadog-agent-integrations.rb
gmmeyer: I believe pip just ignores duplicates in this case

config/software/datadog-agent.rb
gmmeyer: Yes, for the constraints file, since we have to install the agent as part of building the individual packages.

config/software/datadog-agent-integrations.rb
gmmeyer: oh boy you are right I will fix that. Thank you.

resources/datadog-integrations/software.rb.erb
olivielpeau: After discussing this, I vote in favor of not shipping default files in the individual integration packages, it only affects a very small number of integrations on which we update the default files very rarely :)

PR #107, opened 2017-01-10
Adds a software definition for datadog-trace-agent

talwai:
In plain English, the build steps are:
1. Download and install go1.7.1
2. Checkout github.com/DataDog/datadog-trace-agent
3. Download and install datadog-trace-agent build deps
4. Build the datadog-trace-agent with go1.7.1
5. Drop the compiled binary into `install_dir/bin/trace-agent`

This software definition has been tested on builds for:
- [x] deb x86_64
- [x] rpm x86_64
- [x] rpm suse 64


config/projects/datadog-agent.rb
```diff
@@ -224,6 +224,10 @@
   dependency 'datadog-metro'
 end
 
+if linux? and ohai['kernel']['machine'] == 'x86_64'
```
remh: not needed, we are not building i386 anymore

PR #113, opened 2017-01-26
Add support for changing per-user to per machine install

derekwbrown:
Prior versions installed as perUser, which is undesirable.

Set our installation to per machine.
Provide wrapper executable which will uninstall the old (per-user) version and install the new (per-machine) version.  Note that the wrapper executable still needs to be run as the same user as the original install, or the wrapper will fail (but will fail rather than leaving with overlapping installs)


derekwbrown:
@degemer The MSI is still built.  I'll see if I can clean up the bundle UI.

PR #114, opened 2017-01-31
Fixes individual integrations packages

gmmeyer:
I left a few things in the individual integrations packages when I fixed their counterparts in the agent package. This should fix them there, too.


degemer:
For reference: https://github.com/DataDog/dd-agent-omnibus/pull/111

PR #117, opened 2017-02-01
enables building of multiple integrations

gmmeyer:
These changes to the rakefile will enable us to build more than one integration at once.

There are three options now:

1. It will build a single integration
1. It will build multiple specified integrations
1. It will build every integration for that platform into separate packages (it uses ohai to figure out what platform it's on)

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -1,14 +1,19 @@
 require 'json'
+require 'ohai'
+
 PROJECT_DIR='/dd-agent-omnibus'
+
+@ohai = Ohai::System.new.tap { |o| o.all_plugins(%w{platform}) }.data
 
 namespace :agent do
   desc 'Cleanup generated files'
-  task :clean do
+  task :clean do |t|
     puts "Clean up generated files"
     sh "rm -rf /var/cache/omnibus/pkg/*"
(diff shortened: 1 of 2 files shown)


Rakefile.rb
```diff
@@ -68,6 +66,42 @@
...
+  end
+
+  header = erb_header({
+    'name' => "#{integration}",
+    'version' => "#{integration_version}",
+    'build_iteration' => "#{ENV['BUILD_ITERATION']}",
+    'integrations_repo' => "#{ENV['INTEGRATIONS_REPO']}"
+  })
+
+  header = erb_header({
```
truthbk: you don't seem to be using the header's anywhere currently? Am I missing something? I was expecting something like:
```
sh "(echo '#{header}' && cat #{PROJECT_DIR}/resources/datadog-integrations/software.rb.erb) | erb > #{PROJECT_DIR}/config/software/dd-check-#{ENV['INTEGRATION']}-software.rb"
```

gmmeyer: holy shit I accidentally removed that

gmmeyer: I meant to remove a line I'd put in below it that echo'd the erb template to the terminal and removed them both!

gmmeyer:
🎉

PR #119, opened 2017-02-02
[sdk] fix template to use right bin dir.

truthbk:


resources/datadog-integrations/software.rb.erb
```diff
@@ -48,6 +48,10 @@ build do
       install_cmd += " -c #{install_dir}/agent/requirements.txt "
     end
     install_cmd += " -r #{integration_source_folder}/requirements.txt"
-    pip install_cmd
+
+    build_env = {
+      "PATH" => "/#{install_dir}/embedded/bin:#{ENV['PATH']}",
```
truthbk: We could have to add some `ARCH_FLAGS` here as well depending on the architecture.

truthbk:
Not sure what implications this might have for the windows build....

gmmeyer:
@truthbk there are implications but we'll work around those, it'll probably be a conditional

PR #122, opened 2017-02-03
Fix ruby Nil issues

gmmeyer:
Nil in ruby is a bit weird, sometimes.

Here's the issue, if 
```ruby
ENV["USE_INTEGRATION_SDK"].nil? == true
```
then, the following will also, for some reason evaluate to true:
```ruby
ENV["USE_INTEGRATION_SDK"]
```

I do not, really, know why. If I have a hash and I set a variable in that hash to `nil`, it will not perform like this. It's only with the ENV hash. 

My theory is this: the ENV hash is not a real hash, but some other kind of object. And calling something in that object will evaluate to true if that variable exists at all, even if that variable is nil. Therefore, we have to use it like this. 

In addition, we should not set `USE_INTEGRATION_SDK` via substitution in the shell script. However, we will have to rebuild and repush at least the debian container in order for that to work. This will work regardless of whether we rebuild the container, as well. And, it will also work if someone injects a poorly formed variable into the environment, such as `USE_INTEGRATION_SDK==""`, which is something that's very possible with our build environment. 

So, I think we should do both, regardless of whether we have to rebuild the container.

I tried to replicate it in `pry` and found that it only works if you are using `export`:

```shell
/dd-agent-omnibus greg/fix-nil-references ◼ ❯❯❯ DUMMY_VAR='' && pry
[1] pry(main)> ENV["DUMMY_VAR"]
=> nil
[2] pry(main)> if ENV["DUMMY_VAR"]
[2] pry(main)*   print 'a'
[2] pry(main)* end
=> nil
[3] pry(main)>
/dd-agent-omnibus greg/fix-nil-references ◼ ❯❯❯ export DUMMY_VAR='' && pry
[1] pry(main)> ENV["DUMMY_VAR"]
=> ""
[2] pry(main)> if ENV["DUMMY_VAR"]
[2] pry(main)*   print 'a'
[2] pry(main)* end
a=> nil
[3] pry(main)>
```

We don't want to directly check for nil, however. That is way too buggy. Checking if it's equal to the string `"true"` is much better.


config/projects/datadog-agent.rb
```diff
@@ -193,7 +193,7 @@
 dependency 'psutil'
 dependency 'requests'
 
-if !ENV["USE_INTEGRATION_SDK"]
+unless ENV["USE_INTEGRATION_SDK"] == 'true'
```
truthbk: Yeah I think this is probably better than checking for nil... The `truthy` stuff is tricky in ruby.

gmmeyer: yea checking for nil was too buggy, this was much better, I'll update the description and then merge it

PR #123, opened 2017-02-03
[Windows] Support for integration sdk builds

derekwbrown:
Adds templates for msi-specific resource files for omnibus
Rakefile changes for Windows


resources/datadog-integrations/project.rb.erb
```diff
@@ -42,13 +57,34 @@ package :rpm do
...
+  })
+end
 dependency 'preparation'  # creates required build directories
 dependency 'dd-check-<%= name %>-software'
 
 # Checks should require the datadog agent
 runtime_dependency 'datadog-agent'
 
-extends_packages('datadog-agent', 'datadog')
+if ohai['platform'] == "windows"
```
gmmeyer: I think it would be better to use one of the following (I know you're not a rubyist, I do this sometimes too in languages I am more unfamiliar with):

```ruby
unless windows?
  #...
end

if !windows?
  #...
end

if ohai['platform'] != 'windows'
  #...
end

unless ohai['platform'] == 'windows'
  #...
end
```

resources/datadog-integrations/validate_manifest.rb
```diff
@@ -21,3 +21,19 @@ def validate_manifest_0_1_0(manifest_hash)
     raise "manifest.json needs proper fields, currently missing #{field}. Please refer to documentation" unless manifest_hash.key?(field)
   end
 end
+
+def validate_manifest_0_1_2(manifest_hash)
```
gmmeyer: You should remove this, we got rid of versions other than 0.1.0 for now.

resources/datadog-integrations/msi/source.wxs.erb
```diff
@@ -0,0 +1,115 @@
...
+      </Directory>
+    </Directory>
+
+
+    <!-- Delete application when uninstalling -->
+    <!--
+      RemoveFolderEx requires that we "remember" the path for uninstall.
+      Read the path value and set the APPLICATIONROOTDIRECTORY property with the value.
+    -->
+    <Property Id="APPLICATIONROOTDIRECTORY">
```
derekwbrown: We need it for the same reason we need it in the core agent, in case there are files created at runtime (like .pyc).  But, it wasn't functioning properly.  Redid that section so that now the directory is blown away on uninstall, .pycs and all.

resources/datadog-integrations/software.rb.erb
```diff
@@ -4,9 +4,19 @@ require 'json'
...
+  require_relative "../../resources/datadog-integrations/validate_manifest"
+else 
+  require_relative "<%= PROJECT_DIR %>/resources/datadog-integrations/validate_manifest"
+end
+
 integration_source_folder = '/<%= integrations_repo %>/<%= name %>'
-integration_agent_folder = "#{install_dir}/integrations/<%= name %>"
+
+if windows?
+  integration_agent_folder = install_dir
```
derekwbrown: Yes, at least for now (which makes the packaging much more straightforward).

gmmeyer: There are a small predictable number of files outside the install_dir (I can only think of two), so there's no reason not to constrain the installation. We might also be able to do this on linux, but that's a discussion for a different day.

Rakefile.rb
```diff
@@ -1,30 +1,49 @@
 require 'json'
 require 'ohai'
 
-PROJECT_DIR='/dd-agent-omnibus'
-
 @ohai = Ohai::System.new.tap { |o| o.all_plugins(%w{platform}) }.data
 
+def linux?()
```
derekwbrown: No, it errors out (not sure why).

Rakefile.rb
```diff
@@ -1,30 +1,49 @@
...
+end
+
+def windows?()
+  @ohai['platform_family'] == 'windows'
+end
+
+
+if windows?
+  PROJECT_DIR='c:\dd-agent-omnibus'
+  FSROOT="/c/"
```
derekwbrown: It is cygwin syntax, which is why it needs to be used in some places.  The rm command, at least, seems to be a cygwin binary and requires it.

Rakefile.rb
```diff
@@ -1,30 +1,49 @@
...
+    sh "rm -rf #{FSROOT}opt/datadog-agent"
     t.reenable
   end
 
   desc 'Pull the integrations repo'
   task :'pull-integrations' do
     integration_branch = ENV['INTEGRATION_BRANCH'] || 'master'
 
-    sh "rm -rf /#{ENV['INTEGRATIONS_REPO']}"
+    sh "rm -rf #{FSROOT}#{ENV['INTEGRATIONS_REPO']}"
```
derekwbrown: This was the existing code form @gmmeyer .  Note the only change is the FSROOT so it will execute correctly.

Rakefile.rb
```diff
@@ -62,7 +78,9 @@
   desc 'Import signing RPM key'
   task :'import-rpm-key' do
     # If an RPM_SIGNING_PASSPHRASE has been passed, let's import the signing key
-    sh "if [ \"$RPM_SIGNING_PASSPHRASE\" ]; then gpg --import /keys/RPM-SIGNING-KEY.private; fi"
+    if ENV.has_key?('RPM_SIGNING_PASSPHRASE') and not ENV['RPM_SIGNING_PASSPHRASE'].empty?
```
derekwbrown: I just borrowed that from the agent omnibus script...

Rakefile.rb
```diff
@@ -85,20 +103,33 @@ def prepare_and_execute_build(integration, dont_error_on_build: false)
...
     'version' => "#{integration_version}",
     'build_iteration' => "#{ENV['BUILD_ITERATION']}",
-    'integrations_repo' => "#{ENV['INTEGRATIONS_REPO']}"
+    'integrations_repo' => "#{ENV['INTEGRATIONS_REPO']}",
+    'guid' => "#{manifest['guid']}",
+    'description' => "#{manifest['description']}"
   })
 
-  sh "(echo '#{header}' && cat #{PROJECT_DIR}/resources/datadog-integrations/project.rb.erb) | erb > #{PROJECT_DIR}/config/projects/dd-check-#{ENV['INTEGRATION']}.rb"
+  `(echo '#{header}' && cat #{PROJECT_DIR}/resources/datadog-integrations/project.rb.erb) | erb > #{PROJECT_DIR}/config/projects/dd-check-#{integration}.rb`
```
derekwbrown: The sh doesn't process the && syntax properly, the "`" does.

Rakefile.rb
gmmeyer: tbh we probably do not need to rm -rf that. I'm not sure why we do that. I'll investigate, my intuition is that there's not a good reason and that we should just remove it.

Rakefile.rb
gmmeyer: The sh function and the backticks are, somewhat confusingly, not the same.

Rakefile.rb
gmmeyer: It errors out because lib/ostools assumes that it's running in omnibus

gmmeyer: That's also why we have to call ohai, as you see above this line

degemer: Ah, sorry, I forgot about that.

Rakefile.rb
degemer: Ok, it can be part of another PR then, thanks!

Rakefile.rb
degemer: `sh` is a rake DSL if I remember correctly, could a`system` call work instead? (The backticks syntax is not very Ruby-beginner friendly)

gmmeyer: sh does not access system, it accesses a Rake specific system call. (Also we use the backticks all over our stuff, because it returns, unlike sh.)

degemer: Oh yeah, I meant the `system` Ruby syntax - https://ruby-doc.org/core-2.2.0/Kernel.html#method-i-system

PR #126, opened 2017-02-09
[preinst] if find fails, do so silently

degemer:
This line is always annoying when installing the agent:
`find: ‘/opt/datadog-agent/agent’: No such file or directory`
Let's remove it.


degemer:
Not really, I just noticed a few files not ignored, so I added them here. I can remove them if you prefer @truthbk :)

PR #130, opened 2017-02-13
Consolidate dependencies into a single file which can be included by all

derekwbrown:


config/software/agent-deps.rb
```diff
@@ -0,0 +1,57 @@
...
+  dependency 'procps-ng'
+  dependency 'sysstat'
+end
+# Ship supervisor anywhere but on Windows
+if not windows?
+  dependency 'supervisor'
+  dependency 'zlib'
+else
+  # We use our own supervisor shipped as a py2exe-built executable on Windows...
+  # therefore we need py2exe. We also need psutil for our home-made supervisor.
```
degemer: Nit: Could you delete the reference to psutil? It's a core dep now.

config/software/datadog-agent-integrations.rb
```diff
@@ -93,11 +93,16 @@
...
 
   pip_cmd = "install --install-option=\"--install-scripts=#{windows_safe_path(install_dir)}/bin\" -c #{install_dir}/agent/requirements.txt -r /check_requirements.txt"
-  build_env = {
-    "LD_RUN_PATH" => "#{install_dir}/embedded/lib",
-    "PATH" => "/#{install_dir}/embedded/bin:#{ENV['PATH']}",
-  }
-  command "pip #{pip_cmd}", :env => build_env
+  if windows?
+    inst_cmd = "#{windows_safe_path(install_dir)}\\embedded\\scripts\\pip.exe " + pip_cmd
+    command inst_cmd
```
degemer: The `pip` DSL already does that - https://github.com/DataDog/omnibus-ruby/blob/datadog-5.5.0/lib/omnibus/builder.rb#L96-L103, so we can use it (we were using it previously, not sure why it disappeared)

config/software/datadog-agent-integrations.rb
```diff
@@ -93,11 +93,16 @@
...
+  else
+    build_env = {
+      "LD_RUN_PATH" => "#{install_dir}/embedded/lib",
+      "PATH" => "/#{install_dir}/embedded/bin:#{ENV['PATH']}",
+    }
+    command "pip #{pip_cmd}", :env => build_env
```
degemer: Same, `pip` accepts options

config/software/agent-deps.rb
derekwbrown: I tried w/o psutil, and the build failed because it couldn't find psutil.  So I put it back in.

degemer: Ah yes, I was talking of the comment :)

PR #131, opened 2017-02-14
[SDK][templates] we dont need to ship dependencies with core checks 

truthbk:
Dependencies will already be in core by definition. Current state will require bumping the core if there's a library bump, so it's pointless to ship dependencies with the core checks.


olivielpeau:
PR is outdated now that we're using wheels to package individual checks. Closing.

PR #132, opened 2017-02-15
Dynamic Build Instructions

gmmeyer:
Omnibus essentially runs a file twice. Once at the beginning and one at the end. The second time it parses the file, it only runs Omnibus DSL methods.

In order to run arbitrary dynamic ruby, you have to wrap the code in `block do end`. This does that in the integrations and allows it to work. Otherwise, it would only work with an initialized cache.

### DIFF
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -31,6 +31,8 @@ build do
 
   checks = []
 
+  # build do cannot have fully dynamic actions in it
+  # Dynamic actions must be put inside of "block do"
   block do
     # Grab all the checks
     checks = Dir.glob("#{project_dir}/*/")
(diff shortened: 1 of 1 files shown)


config/software/datadog-agent-integrations.rb
```diff
@@ -29,75 +29,84 @@
...
-    all_reqs_file = File.open(all_reqs_file_path, 'w+')
-  else
-    all_reqs_file = File.new(all_reqs_file_path, 'w+')
-  end
+  checks = []
+
+  # build do cannot have fully dynamic actions in it
+  # Dynamic actions must be put inside of "block do"
+  # since most of this is dynamic, I'll wrap the whole thing in `block do`
+  block do
```
truthbk: tricky stuff, great find!

gmmeyer:
The current behavior is this:

An omnibus DSL within the `block do end` will be pushed on the end of the queue and executed after all other DSLs have executed.


Because this is entirely dynamic, that's fine. We just wrap the whole thing in a block. However:

1. For anything else where use this with we will want to be sure that it executes in the proper order every time, under every scenario.
1. We will have to be very careful during an upgrade, to ensure that it continues working even if we upgrade omnibus, because this behavior seems potentially very brittle.

PR #134, opened 2017-02-21
Db/fix agentlaunch

derekwbrown:
[Windows] [Omnibus]  PR fixes multiple issues
-  Automatically closes agent-manager.exe on upgrade/uninstall.  
-  Provide checkbox at the end of the installer to allow user to choose whether to start agent-manager or not (was launching always)
- When executing MSI, still run check for old per-user build, and fail the install (we can't upgrade) in that case. (The EXE will do the proper upgrade).

### DIFF
--- a/resources/datadog-agent/msi/source.wxs.erb
+++ b/resources/datadog-agent/msi/source.wxs.erb
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" 
+     xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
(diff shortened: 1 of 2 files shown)


resources/datadog-agent/msi/source.wxs.erb
```diff
@@ -223,6 +233,12 @@
       <ComponentRef Id="RegisterConfVariables" />
     </Feature>
 
+    <InstallExecuteSequence>
+      <Custom Action="WixCloseApplications"
+          After="StopServices"><![CDATA[REMOVE="ALL" OR REINSTALL]]></Custom>
```
degemer: Out of curiosity, what does that mean? Do you have a reference to build those conditions?

derekwbrown: Which, the "CDATA....?"

degemer: Yes, I found this explanation about the syntax - https://www.firegiant.com/wix/tutorial/com-expression-syntax-miscellanea/expression-syntax/, wondering about the variables

PR #143, opened 2017-02-25
[preinst] Check for proxy and prexisting pgp key

dsalvador-dsalvador:
The current logic breaks on servers with limited connections to the internet.

Check if the key already exists.  Private Apt mirrors can provide these keys.

Alternatively, check for a proxy and apply it to the apt-key command.

### DIFF
--- a/package-scripts/datadog-agent/preinst
+++ b/package-scripts/datadog-agent/preinst
@@ -34,8 +34,8 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
 
     # Prepare the GPG keys rotation:
     # Add the new one to the list of trusted APT keys.
-    # Some servers are on severely resticted networks.  Check for proxy,
-    # and check if key already exists.
+    # Some servers are on severely restricted networks.
+    # Check if key already exists.
     echo "Prepare Datadog Agent keys rotation"
     echo -n "  Add the new 'Datadog, Inc <package@datadoghq.com>' key to the list of APT trusted keys."
     if $( apt-key finger | grep -q 'A292 3DFF 56ED A6E7 6E55  E492 D3A8 0E30 382E 94DE' ); then
--- a/package-scripts/datadog-agent/preinst
+++ b/package-scripts/datadog-agent/preinst
@@ -34,10 +34,18 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
 
     # Prepare the GPG keys rotation:
     # Add the new one to the list of trusted APT keys.
+    # Some servers are on severely resticted networks.  Check for proxy,
+    # and check if key already exists.
     echo "Prepare Datadog Agent keys rotation"
     echo -n "  Add the new 'Datadog, Inc <package@datadoghq.com>' key to the list of APT trusted keys."
-    apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 382E94DE >/dev/null 2>&1 && echo " ... OK" || echo " ... failed"
-
+    if $( apt-key finger | grep -q '935F 5A43 6A5A 6E87 88F0  765B 226A E980 C7A7 DA52' ); then
+        echo "... key already installed"
+    else
+        if [ "$http_proxy" != "" ]; then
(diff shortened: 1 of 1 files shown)


dsalvador-dsalvador:
I accidentally included the fingerprint for the key which expires on 2018-03-04.  New fingerprint is for the key that expires on 2022-06-28.

remh:
Thanks @dsalvador-dsalvador we'll review this shortly

package-scripts/datadog-agent/preinst
```diff
@@ -34,10 +34,18 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
...
+    # and check if key already exists.
     echo "Prepare Datadog Agent keys rotation"
     echo -n "  Add the new 'Datadog, Inc <package@datadoghq.com>' key to the list of APT trusted keys."
-    apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 382E94DE >/dev/null 2>&1 && echo " ... OK" || echo " ... failed"
-
+    if $( apt-key finger | grep -q 'A292 3DFF 56ED A6E7 6E55  E492 D3A8 0E30 382E 94DE' ); then
+        echo "... key already installed"
+    else
+        if [ "$http_proxy" != "" ]; then
+            HKP_PROXY="--keyserver-options http-proxy=$http_proxy"
```
olivielpeau: From `man gpg` in the `--keyserver-options` section:
> http-proxy=value
>     Set the proxy to use for HTTP and HKP keyservers.  This overrides the "http_proxy" environment variable, if any.

So I don't think we need to handle the proxy setting manually here.

I've also tried on a local environment and `apt-key adv` seems to pick up the proxy defined in the `http_proxy` env var on its own.

If you agree let's remove this block

gmmeyer:
Hey @dsalvador-dsalvador can you clarify what you mean when you say the logic breaks? Do you mean the install fails or just that fetching the keys fails? I can replicate failing to fetch the keys, but it always seems to install just fine (sans the keys).

dsalvador-dsalvador:
@gmmeyer The new keys fail to fetch, which causes the chef-datadog cookbook to fail during chef-client runs.

dsalvador-dsalvador:
@olivielpeau Removed the if block that sets the proxy argument.  Although I left the part that checks if the key already exists.

olivielpeau:
@dsalvador-dsalvador Thanks for the clarification. I don't think the present PR will fix this specific issue: if the chef run fails it's probably caused by this resource in the datadog cookbook that would fail on a node that doesn't have the key installed and doesn't have access to the key server: https://github.com/DataDog/chef-datadog/blob/v2.8.1/recipes/repository.rb#L29-L32

One way to work around it would be to install the key before the datadog cookbook's recipes are run. If that doesn't sound like a good solution, could you open an issue against the `chef-datadog` so that we can discuss the issue there?

Even though the present PR won't solve your issue it looks like a good improvement, I'm going to do some local testing to check that it works as expected.

package-scripts/datadog-agent/preinst
```diff
@@ -34,10 +34,15 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
 
     # Prepare the GPG keys rotation:
     # Add the new one to the list of trusted APT keys.
+    # Some servers are on severely resticted networks.  Check for proxy,
+    # and check if key already exists.
```
olivielpeau: One last thing:
- there's a typo `resticted` -> `restricted`
- could you update the comment? (i.e. remove or clarify `Check for proxy` since the script itself doesn't check for proxy anymore :) )

dsalvador-dsalvador:
@olivielpeau That link to the resource isn't where the cookbook fails, it was during the installation of the deb package, which is in the dd-handler recipe before the repository recipe runs.  I like your proposed change to the cookbook, though.  I'm just not sure yet on how to implement it.

olivielpeau:
@dsalvador-dsalvador Merging, thanks!

About the issue on the chef cookbook, I'll let you open an issue on the `chef-datadog` repo if you're still having issues with it and we can discuss it there, thanks

PR #146, opened 2017-03-06
Fixes OS Detection

gmmeyer:
`/etc/issue` is not a great way to detect OS because people edit it sometimes. Therefore, there need to be other ways to detect the OS. This will bring detection in line with how we do os detection in the dd-agent [install_script](https://github.com/DataDog/dd-agent/blob/master/packaging/datadog-agent/source/install_agent.sh#L63-L87).


gmmeyer:
I can also just move it all to the top like we do in the install_script, interested in other people's thoughts on this

gmmeyer:
@degemer I'll make a card to refactor it, but I don't think I have time right this moment

PR #148, opened 2017-03-07
update trace-agent build configuration

galdor:
- Build the trace-agent with rake.
- Make build variables optional.

PR #153, opened 2017-03-09
[rpm][trace] Avoid using `go build -a` on CentOS 5

olivielpeau:
Temp workaround to unblock our rpm build. This build is broken since https://github.com/DataDog/dd-agent-omnibus/pull/148

See https://circleci.com/gh/DataDog/docker-dd-agent-build-rpm-x64/1694 for an example of a build that failed with `go build -a`

cc @galdor: Could you have a look into this and see if there's a better solution?

### DIFF
--- a/config/software/datadog-trace-agent.rb
+++ b/config/software/datadog-trace-agent.rb
@@ -54,6 +54,11 @@ build do
 
    # Build datadog-trace-agent
    command "$GOPATH/bin/glide install", :env => env, :cwd => agent_cache_dir
-   command "rake build", :env => env, :cwd => agent_cache_dir
-   command "mv ./trace-agent #{install_dir}/bin/trace-agent", :env => env, :cwd => agent_cache_dir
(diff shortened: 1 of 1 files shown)


config/software/datadog-trace-agent.rb
```diff
@@ -54,6 +54,11 @@
 
    # Build datadog-trace-agent
    command "$GOPATH/bin/glide install", :env => env, :cwd => agent_cache_dir
-   command "rake build", :env => env, :cwd => agent_cache_dir
-   command "mv ./trace-agent #{install_dir}/bin/trace-agent", :env => env, :cwd => agent_cache_dir
+   if rhel? # temporary workaround for RHEL 5 build issue with the regular `build -a` command
+     command "rake install", :env => env, :cwd => agent_cache_dir
```
truthbk: Will build with `go build -i -o $GOPATH/bin/trace-agent` instead

PR #155, opened 2017-03-10
[python] Extract removal of python-compiled files to separate def

olivielpeau:
This step has to be done at the end of the build to ensure it works well, so it makes sense to extract it. Related PR on omnibus-software: https://github.com/DataDog/omnibus-software/pull/130

Also, move `version-manifest` to the very end of the build.

I've run builds against this branch and made sure that the resulting RPM pkg doesn't output errors on removal anymore


olivielpeau:
cc @hush-hush since you found the issue

PR #161, opened 2017-03-31
build the trace agent with go 1.8

galdor:


config/software/datadog-trace-agent.rb
```diff
@@ -26,7 +26,7 @@
 
 env = {
   "GOPATH" => gopath,
-  "GOROOT" => "/usr/local/go17/go",
+  "GOROOT" => "/usr/local/go18/go",
```
degemer: `"GOROOT" => "#{godir}/go"` to avoid changing it every time? :)

PR #162, opened 2017-03-31
[datadog-agent-integrations] remove install-script

degemer:
Because the psycopg2 upgrade to 2.7.1 broke master.
It is safe for two reasons:
- checks don't (and shouldn't) rely on CLIs to use python modules.
- the default scripts location is #{install_dir}/embedded/bin, and is
  added to the path by the diverse init scripts.

And also: use the pip DSL, it's already cross-platform.


degemer:
Working on Windows.

PR #165, opened 2017-04-11
[datadog-agent] Use option to always fetch tags

olivielpeau:
Related to https://github.com/DataDog/omnibus-ruby/pull/41

Should finally ensure that we always pull the latest tag that's on the remote even when the cache already has the latest commit. And this should allow us to never clear the cache for tag-related reasons.

### DIFF
--- a/config/software/datadog-agent.rb
+++ b/config/software/datadog-agent.rb
@@ -5,7 +5,7 @@ always_build true
 
 local_agent_repo = ENV['LOCAL_AGENT_REPO']
 if local_agent_repo.nil? || local_agent_repo.empty?
-  source git: 'https://github.com/DataDog/dd-agent.git'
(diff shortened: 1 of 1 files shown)


olivielpeau:
(I put this on the `5.12.4` milestone because it'd be great to start using it for `5.12.4` but it's in no way required for `5.12.4`)

olivielpeau:
Discussed with @gmmeyer IRL, the `integrations-core` software def doesn't need this because we don't use the tags of `integrations-core` to determine the version of any omnibus project, and this new option only affects the way omnibus determines what the version of the built software is, not the version that's actually built.

(to put it another way: omnibus has a very reliable way to determine which commit to actually build from, it directly queries the remote to do that. The issue that this option hopefully solves happens at the end of the build when omnibus tries to determine which version it should use for the package: at this point, if the tags haven't been fetched into the local repo, omnibus doesn't apply the correct version to the package)

PR #168, opened 2017-04-25
Whitelists agent files

gmmeyer:
Omnibus doesn't think that the libraries will be there, so it fails the health checks. It shouldn't, though, it should pass them.

### DIFF
--- a/resources/datadog-integrations/software.rb.erb
+++ b/resources/datadog-integrations/software.rb.erb
@@ -8,6 +8,7 @@ if windows?
   require_relative "../../resources/datadog-integrations/validate_manifest"
 else
   require_relative "<%= project_dir %>/resources/datadog-integrations/validate_manifest"
+    whitelist_file /\/opt\/datadog-agent\/.+\.so.+/
 end
 
 integration_source_folder = '/<%= integrations_repo %>/<%= name %>'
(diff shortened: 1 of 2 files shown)


resources/datadog-integrations/software.rb.erb
```diff
@@ -8,6 +8,7 @@ if windows?
   require_relative "../../resources/datadog-integrations/validate_manifest"
 else
   require_relative "<%= project_dir %>/resources/datadog-integrations/validate_manifest"
+  whitelist_file /\/opt\/datadog-agent\/.+\.so.+/
```
olivielpeau: should we only whitelist files that are in `/opt/datadog-agent/embedded`, so that the health check still takes into account the files in `<integration_agent_folder>`?

gmmeyer: I don't think there are any anywhere else, but covering everything wouldn't be harmful, would it?

olivielpeau: Hmm I think the integration-specific deps are in `/opt/datadog-agent/integrations/<name>/lib` so it'd be good to still cover them I think.

AFAIK we only need to whitelist the ones in ``/opt/datadog-agent/embedded` since we assume that the agent build's healthcheck has already taken care of those.

gmmeyer: okay I'll limit it now

PR #169, opened 2017-04-28
[windows] duplicate the windows DLLs into the embedded/DLLs path

derekwbrown:
This PR results in a 3rd copy of the MS runtime DLLs being installed; however, on server 2008, the service fails to start without this (and the other) copies.

### DIFF
--- a/config/software/datadog-agent.rb
+++ b/config/software/datadog-agent.rb
@@ -184,6 +184,13 @@ build do
     # The GUI also needs to have the certificate in its folder to send flares
     copy "datadog-cert.pem", "#{install_dir}/dist/datadog-cert.pem"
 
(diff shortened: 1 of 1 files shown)


config/software/datadog-agent.rb
```diff
@@ -184,6 +184,9 @@
     # The GUI also needs to have the certificate in its folder to send flares
     copy "datadog-cert.pem", "#{install_dir}/dist/datadog-cert.pem"
 
+    copy "#{install_dir}/dist/msvc*.dll", "#{install_dir}/embedded/DLLs"
```
olivielpeau: could you just add a comment above these 2 lines to explain that we need this for windows server 2008?

Will help get rid of these lines when we drop the support of that version of windows

PR #171, opened 2017-05-03
[package-scripts] Fix OS detection for "exotic" distributions

olivielpeau:
Regression introduced in #146

This fixes the name of the variable (`DISTRIBUTION` instead of `LINUX_DISTRIBUTION`) against which tests are performed to determine what the distribution is. Since `LINUX_DISTRIBUTION` was never set, all the comparisons against it were `false`.

The fix should only affect "exotic" distributions for which the tests on the existence of specific distrib files wasn't enough.

The absence of bug reports since #146 has been merged is probably a sign
that these distributions are not widely used ;)


olivielpeau:
Thanks @derekwbrown for the review!

@truthbk This _could_ be included in the next bugfix release, it fixes a regression that's been here since `5.12.0`. That said it doesn't seem to affect that many environments so it's not a must-have for the next bugfix

Issue #177, opened 2017-06-05
Ubuntu/Debian debsums. wrong md5sum ofor /usr/bin/dd-agent


gasiorroiasg:
After proper install of datadog-agent from repository:
`https://apt.datadoghq.com/ stable main`

the file /usr/bin/dd-agent is reported on my hosts as modified.
I looked into that and I've found out that in postInst script 
a /usr/bin/dd-agent file is overwritten with symbolic link:
``
ln -sf $INSTALL_DIR/agent/agent.py `/usr/bin/dd-agent`
`` 
as a result it still has a md5 hash of overwritten file in package md5sums file.
and my periodic ``debsums -c``  check  reports an error on ``/usr/bin/dd-agent`` file

I consider it as a bug of install scripts and it would have been nice to get rid of this.

**Additional environment details (Operating System, Cloud provider, etc):**
i run Debian Wheezy,Debian Jessie hosts. some physical some virtual. 
install process on these don't differ and error still occur.

**Steps to reproduce the issue:**
1. install ``datadog-agent`` package on Debian Jessie host. run```apt-get install datadog-agent```
2. after installation complete run ```debsums -c```

**Additional information you deem important (e.g. issue happens only occasionally):**
The "bug" as i've called it does not affect overall datadog agent behavior or run but in my case i have to consider it as a minor security issue.

PR #178, opened 2017-06-05
Add manualy created md5sums for Debian/Ubuntu package

gasiorroiasg:
At the moment md5sums created while package building are incorrect because of postinst script which is creating symbolic link instead of using original dd-agent file. 
adding md5sums for  this file will check if file (agent.py) has been tampered with or link has been replaced. 
I've could just remove the check for usr/bin/dd-agent but this will expose check for what i mentioned above.


olivielpeau:
Hi @gasiorroiasg, thanks for your PR, and for the detailed explanation! (linking to #177 for more context on the issue)

First off, have you been able to test your change by actually building a package and installing it? :)

I have 2 small concerns/questions about this change:
- the maintainability of this custom md5sums template in the long term (for instance, there's a good chance we forget about its existence and miss it if at some point we change the symlink). We should at least document the existence of this file in https://github.com/DataDog/dd-agent-omnibus/blob/5.14.0/config/projects/datadog-agent.rb#L149. What do you think?
- I'm not sure there's a way to do the same thing with the RPM package, and I'd prefer keeping things as common as possible between the 2 packages... The only way I know of to have a common approach would be to remove `/usr/bin/dd-agent` file from the package file lists entirely. Do you happen to know an approach similar as yours, for the rpm package?

cc @truthbk, who might have ideas/comments on this too

gasiorroiasg:
Hi @olivielpeau 
yes i created the package (via modified Docker container). it is working as i haven't really changed anything just added md5sums file.
regarding bullet-points:
- i will be happy to document this if this piece will be merged.
- RPM is different in this way and therefore adding sums is not needed (maybe by some external script or something). i can't think of any way to add same part for other packages types.
I need this for deb only. and way i did that only affects deb packages. therefore it is only 'patch' type solution.

To maintain more commonness in packages at first it looked like deleting  /usr/bin/dd-agent seems more accurate. but as far as i was able to find out this file is somehow needed in RPMs (there is a note on this in /dd-agent-omnibus/config/projects/datadog-agent.rb)

> Just a dummy file that needs to be in the RPM package list if we don't want it to be removed
   during RPM upgrades. (the old files from the RPM file list that are not in the new RPM file
   list will get removed, that's why we need this one here)
so my only other suggestion (that md5sum piece ) is to remove /usr/bin/dd-agent creation on other package builds and leave it on RPMs only. not sure which is better, or if any is good at all.

gasiorroiasg:
@olivielpeau  Sorry. i've destroyed my master branch.  (amateur hour) (facepalm)
now it is fixed...  i will continue testing different approach on different branch .

djmitche:
It's been a while, so I'm going to close this as stale.  It sounds like work proceeded in another branch, anyway.

PR #180, opened 2017-06-23
[WIP] [macOS] Native GUI

olivielpeau:
**[WIP]**

### What does this PR do?

Replaces the Qt/python/py2app-based macOS GUI build with a native Swift GUI.

See Datadog/dd-agent#3398 for the GUI code

### Motivation

Restore sanity in the macOS build

### Testing

The build works. Still need to double check we ship the tray icon correctly

### Additional notes

For now I've removed `psycopg2` from the mac build, for some reason the omnibus health check fails on it. Will have another look to see what could be the cause


olivielpeau:
This has been included (with minor modifications) in Agent v6, and will not be backported in A5. Closing

PR #181, opened 2017-06-30
Add datadog-process-agent to build

conorbranagan:
Following the release process docs (https://github.com/DataDog/devops/wiki/Process-Agent#dd-agent-release-process) we will be building a statically-linked binary with our own build environment and omnibus will pull it in based on the chosen branch.

cc @shang-wang

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ See:
 ```bash
 PLATFORM="deb-x64" # must be in "deb-x64", "deb-i386", "rpm-x64", "rpm-i386"
 TRACE_AGENT_BRANCH="master" # Branch of the datadog-trace-agent repo to use, default "master"
+PROCESS_AGENT_BRANCH="master" # Branch of the datadog-process-agent repo to use, default "master"
 AGENT_BRANCH="master" # Branch of dd-agent repo to use, default "master"
 OMNIBUS_BRANCH="master" # Branch of dd-agent-omnibus repo to use, default "master"
 AGENT_VERSION="5.4.0" # default to the latest tag on that branch
@@ -45,6 +46,7 @@ docker run --name "dd-agent-build-$PLATFORM" \
   -e AGENT_BRANCH=$AGENT_BRANCH \
   -e AGENT_VERSION=$AGENT_VERSION \
   -e TRACE_AGENT_BRANCH=$TRACE_AGENT_BRANCH \
+  -e PROCESS_AGENT_BRANCH=$PROCESS_AGENT_BRANCH \
   -e RPM_SIGNING_PASSPHRASE=$RPM_SIGNING_PASSPHRASE \
   -e LOCAL_AGENT_REPO=$LOCAL_AGENT_REPO # Only to use if you want to build from a local repo \
   -v `pwd`/pkg:/dd-agent-omnibus/pkg \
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -45,6 +46,7 @@ docker run --name "dd-agent-build-$PLATFORM" \
   -e AGENT_BRANCH=$AGENT_BRANCH \
   -e AGENT_VERSION=$AGENT_VERSION \
   -e TRACE_AGENT_BRANCH=$TRACE_AGENT_BRANCH \
+  -e PROCESS_AGENT_BRANCH=$PROCESS_AGENT_BRANCH \
```
truthbk: you'll have to create PRs on the:
- https://github.com/DataDog/docker-dd-agent-build-deb-x64
- https://github.com/DataDog/docker-dd-agent-build-rpm-x64
- https://github.com/DataDog/docker-dd-agent-build-rpm-suse-x64

on the `circle.yml` file to add support for the flag.

config/projects/datadog-agent.rb
```diff
@@ -177,6 +177,10 @@
   dependency 'datadog-trace-agent'
 end
 
+if linux?
+  dependency 'datadog-process-agent'
```
truthbk: I would just add this below `dependency 'datadog-trace-agent'` above.

config/software/datadog-process-agent.rb
```diff
@@ -0,0 +1,15 @@
+name "datadog-process-agent"
```
truthbk: Nice, you wouldn't have been able to pull from the private repo from circleCI (while building) anyways, so this is the way to go. It'll also keep the build speedy.

config/software/datadog-process-agent.rb
```diff
@@ -0,0 +1,15 @@
...
+always_build true
+
+build do
+  agent_branch = ENV['PROCESS_AGENT_BRANCH']
```
truthbk: Also, I actually skimmed over this way too quickly earlier. I would probably move this out of the `build` block, and set the agent_branch there. As well as setting the `default_version` - I don't think it's necessary, but it'll help be consistent with other definitions.

config/software/datadog-process-agent.rb
```diff
@@ -0,0 +1,15 @@
...
+  if agent_branch.nil? || agent_branch.empty?
+    agent_branch = "master"
+  end
+
+  binary = "process-agent-amd64-#{agent_branch}"
```
truthbk: If you set `default_version` (expectedly `default_version agent_branch` immediately after collecting `agent_branch`) you can then use the `version` function here, it'd be a little more readable.

truthbk: And you probably want to use the `ship_license` omnibus method to make sure we ship the corresponding license.

conorbranagan: Can't ship a license until the repo is open source (I think?) so I'll add a FIXME for now so that we can fix that before the next release.

truthbk:
@conorbranagan It's too late now... but rewriting my `dogstatsd6` definition I noticed you should probably be using the omnibus file fetcher, instead of using `curl` explicitly. See this for an example: https://github.com/DataDog/omnibus-software/blob/master/config/software/jmxfetch.rb

PR #184, opened 2017-07-13
do not install pip deps for some integrations

gmmeyer:
This should fix the integrations issue. It is not a great solution: I would vastly prefer to be able to do ship them with upgraded dependencies. However, this is the solution we have. So, as sad as I am at the moment to give up the dream of shipping updated deps, it's all we can do. 😢

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -45,7 +45,7 @@ namespace :agent do
     # if someone forgets to set INTEGRATIONS_REPO
     raise 'INTEGRATIONS_REPO not set!' unless ENV['INTEGRATIONS_REPO']
     sh "rm -rf #{FSROOT}#{ENV['INTEGRATIONS_REPO']}"
-    
+
     sh "git clone https://github.com/DataDog/#{ENV['INTEGRATIONS_REPO']}.git /#{ENV['INTEGRATIONS_REPO']} || true"
     sh "cd /#{ENV['INTEGRATIONS_REPO']} && git checkout #{integration_branch}"
     sh "cd /#{ENV['INTEGRATIONS_REPO']} && git fetch --all"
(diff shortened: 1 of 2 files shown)


resources/datadog-integrations/software.rb.erb
```diff
@@ -52,27 +52,28 @@ build do
...
-
   if File.exists? "#{integration_source_folder}/requirements.txt"
-    # install requirements
-    install_cmd = "install --target=#{integration_agent_folder}/lib "
-    # Only apply the constraints file if one exists
-    if File.exists? "#{install_dir}/agent/requirements.txt"
-      install_cmd += " -c #{install_dir}/agent/requirements.txt "
-    end
-    install_cmd += " -r #{integration_source_folder}/requirements.txt"
+    unless manifest.key?("package_deps") && ! manifest["package_deps"]
```
truthbk: Wouldn't this be easier to read with an `if` instead?

Rakefile.rb
```diff
@@ -71,11 +71,18 @@ def windows?()
   desc 'Build all integrations'
   task :'build-all-integrations' do
     checks = Dir.glob("/#{ENV['INTEGRATIONS_REPO']}/*/")
+    if ENV['SKIP_INTEGRATION']
+      skip_checks = ENV['SKIP_INTEGRATION'].split(',')
```
truthbk: would it be better to split and strip?

```
skip_checks = ENV['SKIP_INTEGRATION'].split(",").map(&:strip)
```

resources/datadog-integrations/software.rb.erb
gmmeyer: I can change it to an if if you'd like!

gmmeyer: I agree that it's clearer, the original was definitely unclear

Rakefile.rb
gmmeyer: Yes, thats a good idea

PR #187, opened 2017-07-27
Remove not needed file from Debian package creationg process

gasiorroiasg:
This piece according to comment should be in RPM related block. Existence of this file on Debian it is not needed, and even more it creates a problem i wanted to solve for some time. 
to act with caution i moved it to a block negating debian distro. therefore if some other distro was needing that it is still there.


olivielpeau:
Closing this PR as I'm not sure it's relevant anymore, and modifies Agent v5 while the more recent v6 and v7 are now available.

PR #188, opened 2017-08-07
Adding gnupg and dirmngr to allow install on debian 9

hush-hush:
Linked to: https://github.com/DataDog/dd-agent/issues/3397

Debian 9 is missing by default packages to use `apt-key`.

### DIFF
--- a/config/projects/datadog-agent.rb
+++ b/config/projects/datadog-agent.rb
@@ -109,6 +109,9 @@ elsif debian?
   replace 'datadog-agent-base (<< 5.0.0)'
   replace 'datadog-agent-lib (<< 5.0.0)'
   conflict 'datadog-agent-base (<< 5.0.0)'
+  # needed starting debian 9
+  runtime_dependency 'gnupg'
+  runtime_dependency 'dirmngr'
 end
 
 # ------------------------------------


olivielpeau:
@hush-hush I have one concern with this change: on ubuntu 14.04 `dirmngr` is provided by the `universe` repository, which may not be enabled on some environments. In that case the install of the agent would fail even though `dirmngr` is not required on that version of ubuntu... (for reference see https://packages.ubuntu.com/trusty/dirmngr)

Since we build one package for all supported versions of ubuntu and debian we can't make the `runtime_depencency` statement depend on the version of the target platform. So if we want to avoid this issue I think we can:
1. remove the `runtime_dependency` on `dirmngr` here
2. document in the error message here: https://github.com/DataDog/dd-agent-omnibus/blob/5.16.0/package-scripts/datadog-agent/preinst#L45 that the `dirmngr` package is needed on debian >= 9

What do you think?

hush-hush:
Good catch !

It's a difficult question: I don't think we should failed the install on debian9 because a package is not available on ubuntu (and not failed either on ubuntu because of debian).

Knowing that ubuntu 14.04 is the most used linux distribution right now I don't see any "good" solution. Maybe we should split the debian and ubuntu package ?

PR #191, opened 2017-08-08
Adding gnupg to allow install on debian 9 and a warning if dirmngr is not installed

hush-hush:
We don't enforce dependency to dirmngr on the .deb package since it's
shared between ubuntu and debian and it's only available in the
'universe' repo in ubuntu <= 14.04.

Packaging for debian and ubuntu will be split in future version of the
agent.

see: https://github.com/DataDog/dd-agent/issues/3397

### DIFF
--- a/package-scripts/datadog-agent/preinst
+++ b/package-scripts/datadog-agent/preinst
@@ -42,7 +42,8 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
     if $( apt-key finger | grep -q 'A292 3DFF 56ED A6E7 6E55  E492 D3A8 0E30 382E 94DE' ); then
         echo "... key already installed"
     else
-        apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 382E94DE >/dev/null 2>&1 && echo " ... OK" || echo " ... failed"
+        apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 382E94DE >/dev/null 2>&1 && echo " ... OK" \
+            ||  printf " ... failed\nWARNING: If you are running Debian 9, the \"dirmngr\" package is needed to install the new APT key required by future versions of the Agent.\nIf this package is not installed, please install it and then re-install the agent. The agent installation may break in the future if the new APT key is not installed on your system\n"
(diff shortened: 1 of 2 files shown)


package-scripts/datadog-agent/preinst
```diff
@@ -42,7 +42,15 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
     if $( apt-key finger | grep -q 'A292 3DFF 56ED A6E7 6E55  E492 D3A8 0E30 382E 94DE' ); then
         echo "... key already installed"
     else
-        apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 382E94DE >/dev/null 2>&1 && echo " ... OK" || echo " ... failed"
+        apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 382E94DE >/dev/null 2>&1
+        if [ $? -eq 0 ]; then
+            echo " ... OK"
+        else
+            echo " ... failed"
+            if [ -f "/etc/debian_version" ] || [ "$DISTRIBUTION" = "Debian" ]; then
```
olivielpeau: since `/etc/debian_version` is also present on Ubuntu, this condition will also be true on Ubuntu.

I think we can either:
1. only use the `$DISTRIBUTION` value, which should work as long as `lsb_release` is actually present on the system (we can enforce that by declaring `lsb-release` as a dependency of the package), or
2. change the error message a bit, to something like: `If you're running Debian 9, [...]`

Honestly I think option `2.` is enough :)

package-scripts/datadog-agent/preinst
```diff
@@ -42,7 +42,15 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
...
+                 echo "The dirmngr package is needed to add a new key to APT on Debian since version 9"
```
olivielpeau: What do you think about adding more details here on the consequences of failing to add the new APT key. Something like: `WARNING: The "dirmngr" package is needed to install the new APT key required by future versions of the Agent. If this package is not installed, please install it and then re-install the agent. The agent installation may break in the future if the new APT key is not installed on your system`

PR #194, opened 2017-08-17
[centos] Improve systemd and service restart behaviors

olivielpeau:
Depends on https://github.com/DataDog/dd-agent/pull/3477

* Improves service restart logic in installation scripts: on CentOS instead of restarting the agent service both in `postinst` and `posttrans`, only restart it in `posttrans`, and make sure a `datadog.conf` file is present before doing so.

* Adds a systemd unit file: CentOS versions that use systemd (>= CentOS 7) will use it instead of the `init.d` script for lifecycle commands (start/stop/restart): same unit file and bash script as what's already used on debian.
  On a customer's environment with CentOS 7, using a systemd unit file has been way more reliable than relying on systemd's conversion of the `init.d` script, so let's ship it with the regular RPM pkg. 

These changes pass `dd-agent-testing`. I haven't done much manual testing but if someone feels we should I can do that.

### DIFF
--- a/config/projects/datadog-agent.rb
+++ b/config/projects/datadog-agent.rb
@@ -125,9 +125,9 @@ if linux?
     extra_package_file '/lib/systemd/system/datadog-agent.service'
   end
 
-  # SysVInit service file
   if redhat?
     extra_package_file '/etc/rc.d/init.d/datadog-agent'
+    extra_package_file '/usr/lib/systemd/system/datadog-agent.service'
   end
 
   if suse?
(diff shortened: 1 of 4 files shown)


olivielpeau:
Just tested this on a CentOS 5 VM, it works well 👍

truthbk:
[[packaging][centos] Improve service (systemd + init script)](https://trello.com/c/rholci6c/751-packagingcentos-improve-service-systemd-init-script)

PR #198, opened 2017-09-14
[Logs] Embed logs agent in dd-agent when enabled

tmichelet:
Include logs-agent in dd-agent builds.

It currently supports linux only (windows should come in the following weeks).

Follow full story: https://github.com/DataDog/dd-agent/pull/3520

### DIFF
--- a/config/software/datadog-logs-agent.rb
+++ b/config/software/datadog-logs-agent.rb
@@ -1,0 +1,14 @@
+name "datadog-logs-agent"
+always_build true
+
+build do
+  binary = "logagent"
+  logs_agent_version = "alpha"
+
+  if linux?
+    url = "https://s3.amazonaws.com/public.binaries.sheepdog.datad0g.com/agent/#{logs_agent_version}/linux-amd64/#{binary}"
+    command "curl #{url} -o #{binary}"
+    command "chmod +x #{binary}"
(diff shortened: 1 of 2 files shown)


config/software/datadog-logs-agent.rb
```diff
@@ -0,0 +1,14 @@
+name "datadog-logs-agent"
+always_build true
+
+build do
+  binary = "logagent"
+  logs_agent_version = "alpha"
```
truthbk: Feel free to ignore if you think it doesn't appy: 
I know you're still in alpha stage - but you might want to prepare the software definition for versioning (I presume you guys will iterate quickly so it might come in handy). You can take a look at the [datadog-process-agent](https://github.com/DataDog/dd-agent-omnibus/blob/master/config/software/datadog-process-agent.rb#L4-L13) software definition for that.

truthbk:
Almost forgot - let's not merge this into `master`, I don't think we want to ship the alpha logger by default :) How about another `5.17.x-logbeta` branch here?

config/software/datadog-logs-agent.rb
tmichelet: This requires some boilerplate work to pass all parameters from jenkins to containers. I thought we could avoid this for now

truthbk:
Not going to merge.

PR #201, opened 2017-09-30
Nick/restart on cent os7

nmuesch:
Enabled the datadog agent service when running with systemd. 

Currently CentOS 7 is using systemd, but the we don't enable the service on the post install.


package-scripts/datadog-agent/postinst
```diff
@@ -82,6 +84,10 @@ if [ "$DISTRIBUTION" != "Darwin" ]; then
       chkconfig --add datadog-agent
   fi
 
+  if command -v systemctl >/dev/null 2>&1; then
+      systemctl enable $SERVICE_NAME
```
olivielpeau: nit: can you simply use `datadog-agent` directly here instead of using a variable? (if we use a variable, we should for consistency make sure that all occurences of the service name are replaced with the variable value)

PR #205, opened 2017-11-22
failure to get process agent will make the build fail

gmmeyer:
Currently, if we fail to grab the process agent, it won't fail. Curl doesn't error out if you get a 4xx or 5xx response code. However, it will if you pass it the option `-f`. This will make it fail noisily.


olivielpeau:
btw this also makes sense for the agent 6 build: I've opened https://github.com/DataDog/datadog-agent/pull/835 that does the same thing there

PR #208, opened 2017-12-18
Add a release.json file containing every information needed about a release

hush-hush:
To monimize the changes we continue to export the information in the
release.json. Also the latest setting were added to release.json to initialize the file.

Side change:
Fix test to work with the latest manifest version from integrations-core


omnibus_build.sh
```diff
@@ -30,6 +24,22 @@ rm -rf /opt/$PROJECT_NAME/*
...
 builtin cd $PROJECT_DIR
 
+# Install the gems we need, with stubs in bin/
+bundle update # Make sure to update to the latest version of omnibus-software
+
+# Loading release information from release.json
+vars=`bundle exec ruby load_realease.rb`
+
+# Export variables from the release.json
+IFS="
```
truthbk: nit: I think I prefer
```
IFS=$'\n'
```
because it's more explicit (but feel free to ignore)

Also, although it doesn't affect the script in the following lines, it might be the right thing to restore IFS to the default behavior:
```
IFS_DEFAULT=$IFS
IFS=$'\n'
for line in $vars;
do
...
done
IFS=$IFS_DEFAULT
```

omnibus_build.sh
```diff
@@ -30,6 +24,22 @@ rm -rf /opt/$PROJECT_NAME/*
 
 builtin cd $PROJECT_DIR
 
+# Install the gems we need, with stubs in bin/
+bundle update # Make sure to update to the latest version of omnibus-software
+
+# Loading release information from release.json
+vars=`bundle exec ruby load_realease.rb`
```
truthbk: huge annoying nitpick: I prefer to stay away from backticks 😊 
```
vars=$(bundle exec ruby load_realease.rb)
```

omnibus_build.sh
```diff
@@ -30,11 +24,26 @@ rm -rf /opt/$PROJECT_NAME/*
...
+
+# Export variables from the release.json
+IFS=$'\n'; for line in $vars
+do
+  key=`echo $line | cut -d ' ' -f 1`
+  value=`echo $line | cut -d ' ' -f 2`
+  echo "setting $key => $value"
+  export "$key=$value"
+done
+IFS=$IFS_DEFAULT
```
truthbk: We're not saving the original `IFS_DEFAULT=$IFS` before the for loop.

Gemfile
```diff
@@ -1,6 +1,9 @@
 source 'https://rubygems.org'
-gem 'omnibus', git: 'git://github.com/datadog/omnibus-ruby.git', branch: ENV['OMNIBUS_RUBY_BRANCH']
-gem 'omnibus-software', git: 'git://github.com/datadog/omnibus-software.git', branch: ENV['OMNIBUS_SOFTWARE_BRANCH']
+
+# we default if env variable aren't set or set to an empty string
```
truthbk: This is fine I guess... I do feel it may be a bit too complex for a gemfile. But I guess we don't have a ton of alternatives

PR #213, opened 2018-01-17
[wheels] bundle integrations as wheels

truthbk:
This PR will bundle the integrations as wheels. 

We first build and then install the built integration's wheels, this adds a little to the build-time, but it shouldn't be too bad. We also compute a constraints file for the agent python dependencies to avoid upgrading any of the core dependencies if any of the wheels (for whatever reason, though it shouldn't be the case) is not aligned with the core agent.

Note:
SQLServer is actually also a linux check, and found `unixodbc` was missing, that is provided in the following omnibus-software PR: https://github.com/DataDog/omnibus-software/pull/176


release.json
```diff
@@ -10,6 +10,17 @@
 		"OMNIBUS_RUBY_BRANCH": "datadog-5.5.0",
 		"OMNIBUS_SOFTWARE_BRANCH": "master"
 	},
+	"5.22.0-wheels": {
```
gmmeyer: Did you mean to leave this here?

config/software/datadog-agent-integrations.rb
```diff
@@ -4,7 +4,8 @@
 
 dependency 'pip'
 dependency 'datadog-agent'
-dependency 'integration-deps'
+
+dependency 'unixodbc'
```
gmmeyer: this should only be installed on linux

PR #217, opened 2018-02-02
[Windows] Add process agent support to agent5.

derekwbrown:
Install both APM and PRocess as an automatic start service (instead of
manual start).  Relies on the services self-stopping if they're so
configured.

Change all services to automatic restart on failure.

Move service names, descriptions, etc. into the localization file

Companion PR https://github.com/DataDog/dd-agent/pull/3655

### DIFF
--- a/resources/datadog-agent/msi/localization-en-us.wxl.erb
+++ b/resources/datadog-agent/msi/localization-en-us.wxl.erb
@@ -1,8 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <WixLocalization Culture="en-us" xmlns="http://schemas.microsoft.com/wix/2006/localization">
   <String Id="LANG">1033</String>
-  <String Id="ProductName"><%= friendly_name %></String>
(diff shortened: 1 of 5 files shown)


resources/datadog-agent/msi/localization-en-us.wxl.erb
```diff
@@ -63,4 +67,15 @@
   <String Id="WillInstall">[WixBundleName] will be installed on your device</String>
   <String Id="InstallLicenseLinkText">By installing you accept these &lt;a href="#"&gt;license terms&lt;/a&gt;</String>
 
+  <!-- these strings are for the main agent service install -->
+  <String Id="ServiceDisplayName">[ProductName]</String>
+  <String Id="ServiceDescription">Send metrics to [ShortCompanyName]</String>
```
masci: `CompanyNameShort` ?

derekwbrown: ShortCompanyName (now) property defined in source.wxs.erb.

PR #227, opened 2018-02-07
[unix] Remove /opt/datadog-agent/agent/checks.d directory

olivielpeau:
We're not shipping anything in it anymore (wheel-based build), let's
remove it to avoid confusion.

### DIFF
--- a/config/software/datadog-agent.rb
+++ b/config/software/datadog-agent.rb
@@ -28,9 +28,6 @@ build do
   mkdir  "#{install_dir}/agent/"
 
   # Agent code
-  mkdir "#{install_dir}/agent/checks.d"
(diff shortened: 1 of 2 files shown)


config/software/datadog-agent.rb
```diff
@@ -28,9 +28,6 @@
   mkdir  "#{install_dir}/agent/"
 
   # Agent code
-  mkdir "#{install_dir}/agent/checks.d"
-  copy 'checks.d', "#{install_dir}/agent/"
```
olivielpeau: just a note: this uses the `checks.d` directory that used to be in the `dd-agent` repo, which doesn't exist since we use `integrations-core`

PR #238, opened 2018-03-06
[release] adding 5.22.2 release

truthbk:


release.json
```diff
@@ -10,6 +10,17 @@
 		"OMNIBUS_RUBY_BRANCH": "datadog-5.5.0",
 		"OMNIBUS_SOFTWARE_BRANCH": "master"
 	},
+	"5.22.2": {
+		"AGENT_VERSION": "5.22.2",
+		"AGENT_BRANCH": "5.22.2",
+		"AGENT6_BRANCH": "6.0.1",
+		"INTEGRATIONS_CORE_BRANCH": "5.22.2",
+		"TRACE_AGENT_BRANCH": "5.22.0",
```
LotharSee: Let's move to `6.0.1` for the `trace-agent`.
Good to you?
I don't know how we should tag/version it while we still have A5 and A6.

truthbk: Normally A5 and A6 releases should be chronologically aligned, so I think continuing with `6.x` tags should be fine. Also if for once we're not strict about semver it's not the end of the world.

LotharSee: Okidoki so let's go with this one 👍

release.json
```diff
@@ -10,6 +10,17 @@
...
 		"OMNIBUS_SOFTWARE_BRANCH": "master"
 	},
+	"5.22.2": {
+		"AGENT_VERSION": "5.22.2",
+		"AGENT_BRANCH": "5.22.2",
+		"AGENT6_BRANCH": "6.0.1",
+		"INTEGRATIONS_CORE_BRANCH": "5.22.2",
+		"TRACE_AGENT_BRANCH": "6.0.1",
+		"TRACE_AGENT_ADD_BUILD_VARS": "true",
+		"PROCESS_AGENT_BRANCH": "5.22.1",
```
sunhay: `5.22.2` for process-agent. Just built the binary now.

sunhay: 🙏

PR #247, opened 2018-04-05
Maxime/set version in omnibus

hush-hush:
Update the `AGENT_VERSION` using release.json.


hush-hush:
- The logic for windows is coming soon. I wanted to hear some feedback first
- The install script is another issue in itself. RIght now we need to update both version when we close a release. But they're not linked to each other. I was thinking that first the install script could update the version in `config.py` and later we could automate the update on `DEFAULT_AGENT_VERSION` with a jenkins jobs. what do you think ?
- I'll update the defautl to `5.99.99`
- I think the good side is that we will be able to see if a customer install a version directly from git (a version that may be in-between two release).

The other solution if we don't want to go that way is to have a job that would update and commit the version in `config.py`, the install script and the `release.json`. That might be simpler in the end if we don't want to have 5.99.99 version in the wild.

PR #249, opened 2018-04-11
[trace-agent] we should be fine to pass the default_version to Makefile

truthbk:
This was going to be prone to set an incorrect version for the trace-agent when building on windows. Let's use `version` instead and let Makefile handle setting the default `0.99.0` for nightlies/untagged versions. 

Sister PR: https://github.com/DataDog/datadog-trace-agent/pull/409


truthbk:
Actually, I think we should probably do this instead:

https://github.com/DataDog/datadog-agent/blob/master/omnibus/config/software/datadog-trace-agent.rb#L53

truthbk:
Actually no need to merge this the PR should be enough to handle the cases where the `AGENT_VERSION` for whatever reason isn't specified in the build.

PR #250, opened 2018-04-12
[datadog-pip] adding tuf+in-toto modified pip

truthbk:
### What does this PR do?

This PR adds the TUF-hardened pip to the agent5 build. We can probably ship this as-is, when invoked plainly, it just breaks like good-old classic pip. As a matter of fact all subsequent integrations wheels are installed with the TUF-pip. 

### Motivation

Future secure distribution of integrations as wheels.


olivielpeau:
@truthbk let's close this one, tuf-hardened pip will not be added to agent 5

truthbk:
Closing!

PR #254, opened 2018-06-05
Change sub services (process & apm) to demand start so they don't

derekwbrown:
fill the event log starting/stopping when not configured


derekwbrown:
@olivielpeau this matches the behavior on A6.  However, IIRC, the default behavior is supposed to be that APM starts by default (unless explicitly disabled), process does not start by default (must be explicitly enabled).

olivielpeau:
ok thanks, you've opened https://github.com/DataDog/dd-agent/pull/3749 to address the behavior of trace-agent

PR #268, opened 2018-07-19
Clean up integration core deps and yaml files

nmuesch:
Changes the locations of the *.yaml files to be pulled from their new location

Changes the dependency handling since we are moving to a static environment. We now install each integration with --no-deps, run a pip-compile on the agent_requirements.in file from datadog_checks_base that list the requirements for all integration and install the reqs of that compiled file.

Remove some seemingly dead code and implement the blacklist for skipping installing non-checks

We were previously doing `pip wheel --no-deps` and then pip install on that wheel. This actually causes the dependencies to still be installed, it just doesn't build the wheel for each dependency. What we want is to actually skip the deps during install, not build. 

Added a dep on pip-tools. We now compile the requirements in the builder instead of relying on them to be in the integration core repo. One benefit here is that we can use environment markers (running pip-compile strips the dep from the output txt if the marker requirement isn't met, so building in the runner allows us to dynamically figure out each dependency for the platform)

### DIFF
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -9,7 +9,7 @@ relative_path 'integrations-core'
 
 PIPTOOLS_VERSION = "2.0.2"
 WHEELS_VERSION = "0.30.0"
-UNINSTALL_PIPTOOLS_DEPS = ['first', 'click', 'pip-tools']
+UNINSTALL_PIPTOOLS_DEPS = ['pip-tools']
 
 # The only integrations that will be packaged with the agent
 # are the ones that are officiallly supported.
@@ -82,7 +82,7 @@ build do
     # Windows pip workaround to support globs
     python_bin = "\"#{windows_safe_path(install_dir)}\\embedded\\python.exe\""
     python_pip_no_deps = "pip install --no-deps #{windows_safe_path(project_dir)}"
-    python_pip_reqs = "pip install -c #{windows_safe_path(install_dir)}\\agent_requirements.txt --require-hashes -r #{windows_safe_path(project_dir)}"
+    python_pip_reqs = "pip install -c #{windows_safe_path(install_dir)}\\agent_requirements.txt --require-hashes -r"
     python_pip_uninstall = "pip uninstall -y"
 
     # Install the static environment requirements that the Agent and all checks will use
@@ -95,7 +95,7 @@ build do
         command("#{python_bin} -m #{python_pip_uninstall} #{dep}")
       end
 
-      command("#{python_bin} -m #{python_pip_reqs}\\static_requirements.txt")
+      command("#{python_bin} -m #{python_pip_reqs} #{windows_safe_path(project_dir)}\\static_requirements.txt")
     else
       build_env = {
         "LD_RUN_PATH" => "#{install_dir}/embedded/lib",
@@ -155,6 +155,10 @@ build do
       end
 
       File.file?("#{project_dir}/#{check}/setup.py") || next
+
+      # Installing each integration with no deps because each integration depends
+      # only datadog_checks_base. All integrations-core requirements are now in requirements.in in the repo
+      # We won't install deps here to help ensure that no new dependencies sneak into the setup.py during the PR review process. 
       if windows?
         command("#{python_bin} -m #{python_pip_no_deps}\\#{check}")
       else
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -73,7 +73,7 @@ build do
     # loop through checks and install each without their dependencies
     # we rely on a static Agent environment that was built above. 
     checks.each do |check|
-      if blacklist.include? check:
+      if blacklist.include?(check):
         next
 
       # Only use the parts of the filename we need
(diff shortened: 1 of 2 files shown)


config/software/datadog-agent-integrations.rb
```diff
@@ -56,18 +56,21 @@
     python_bin = "\"#{windows_safe_path(install_dir)}\\embedded\\python.exe\""
     python_pip_no_deps = "pip install --no-deps #{windows_safe_path(project_dir)}"
     python_pip_reqs = "pip install --require-hashes -r #{windows_safe_path(project_dir)}"
+    python_pip_compile = "pip-compile #{windows_safe_path(project_dir)}"
 
     # Install the static environment requirements that the Agent and all checks will use
     if windows?
       command("#{python_bin} -m #{python_pip_no_deps}\\datadog_checks_base")
-      command("#{python_bin} -m #{python_pip_reqs}\\datadog_checks_base\\datadog_checks\\data\\agent_requirements.txt")
+      command("#{pip_compile}\\datadog_checks_base\\requirements.in > #{windows_safe_path(project_dir)}\\agent_requirements.txt"
```
gmmeyer: should be `python_pip_compile`

config/software/datadog-agent-integrations.rb
```diff
@@ -50,45 +52,43 @@
...
-    else
-      build_env = {
-        "LD_RUN_PATH" => "#{install_dir}/embedded/lib",
-        "PATH" => "#{install_dir}/embedded/bin:#{ENV['PATH']}",
-      }
-      pip "install -r #{project_dir}/check_requirements.txt", :env => build_env
-    end
-
-    # Set frozen requirements
-    pip "freeze > #{install_dir}/agent_requirements.txt"
```
truthbk: We still need to build `agent_requirements` even if we no longer explicitly install the `wheel` package. We need to create this *after* installing pip tools.

The `"{windows_safe_path(project_dir)}\\datadog_checks_base\\agent_requirements.in` isn't the same thing we need the constraints file to ensure the omnibus deps for agent5 are respected above all things. The problem this has had for a long time is that pip will enforce the constraints but will not fail installing a package that doesn't meet the requirements. I know that sucks, ideally we'd want to fail the build :(

config/software/datadog-agent-integrations.rb
```diff
@@ -50,45 +52,43 @@
...
+  # Install all the requirements
+  if windows?
+    pip "install pip-tools==2.0.2"
```
truthbk: Can you guys use a variable up top to pin to `2.0.2`, and use string interpolation to enforce it? Just one place to define this version, please.

config/software/datadog-agent-integrations.rb
```diff
@@ -50,45 +52,43 @@
...
     python_bin = "\"#{windows_safe_path(install_dir)}\\embedded\\python.exe\""
-    python_pip = "pip install -c #{windows_safe_path(install_dir)}\\agent_requirements.txt #{windows_safe_path(project_dir)}"
+    python_pip_no_deps = "pip install --no-deps #{windows_safe_path(project_dir)}"
+    python_pip_reqs = "pip install --require-hashes -r #{windows_safe_path(project_dir)}"
 
+
+    # Install the static environment requirements that the Agent and all checks will use
     if windows?
-      command("#{python_bin} -m #{python_pip}\\datadog_checks_base")
+      command("#{python_bin} -m #{python_pip_no_deps}\\datadog_checks_base")
```
truthbk: why `--no-deps` here?

config/software/datadog-agent-integrations.rb
```diff
@@ -111,31 +111,30 @@
...
-          copy "#{project_dir}/#{check}/auto_conf.yaml", "#{conf_directory}/auto_conf/#{check}.yaml"
+        if File.exists? "#{project_dir}/#{check}/datadog_checks/#{check}/data/auto_conf.yaml"
+          copy "#{project_dir}/#{check}/datadog_checks/#{check}/data/auto_conf.yaml", "#{conf_directory}/auto_conf/#{check}.yaml"
         end
       end
 
       File.file?("#{project_dir}/#{check}/setup.py") || next
       if windows?
-        command("#{python_bin} -m #{python_pip}\\#{check}")
+        command("#{python_bin} -m #{python_pip_no_deps}\\#{check}")
```
truthbk: ditto re:`--no-deps`?

config/software/datadog-agent-integrations.rb
nmuesch: Ok thats good to know. Thanks for the info! I put back the freeze and the use of it as a constraints file. I agree it would be better to fail the build but there isn't a way to do that currently.

config/software/datadog-agent-integrations.rb
nmuesch: pinned!

config/software/datadog-agent-integrations.rb
nmuesch: As discussed offline, all the deps are now in the requirements.in file in each integration but the one that gets used is the big requirement.in file in datadog_checks_base. So we don't need to install the dependencies for a single integration anymore.

nmuesch:
Currently there is an issue with reading the constraints file @truthbk File.read can never find the file despite being there. Originally looked with @ofek  thought it was a symlink issue but it turned out to not be the case. I reverted that change to unblock things but we can make a card to take a look to do this.

PR #270, opened 2018-07-24
[integrations] safer uninstall of build-time dependencies

truthbk:


config/software/datadog-agent-integrations.rb
```diff
@@ -89,29 +98,34 @@
...
       pip "install --no-deps .", :env => build_env, :cwd => "#{project_dir}/datadog_checks_base"
       command("#{install_dir}/embedded/bin/python -m piptools compile --generate-hashes --output-file #{project_dir}/static_requirements.txt #{project_dir}/datadog_checks_base/datadog_checks/data/agent_requirements.in")
+    end
 
-      # Uninstall the deps that pip-compile installs so we don't include them in the final artifact
-      for dep in UNINSTALL_PIPTOOLS_DEPS
-        pip "uninstall -y #{dep}"
+    # Uninstall the deps that pip-compile installs so we don't include them in the final artifact
+    UNINSTALL_PIPTOOLS_DEPS.each do |dep|
+      if not frozen_agent_reqs.include? "#{dep}=="
```
arbll: If pip-tools includes `dummy==1.0` and we want to ship `toto-dummy==2.6`, `toto-dummy==2.6` will get removed

truthbk: No, that's not accurate. We'd *NOT* remove it, which is a lesser problem, but I'll change this for  a regex match.

arbll: Oh yeah you are right I don't know why but I assumed that we were uninstalling what was matched and not `dep`...
Sorry for the noise. I think we should still fix it in case we modify this logic in the future.
I would `include` to 
https://www.rubydoc.info/stdlib/core/1.9.2/String#start_with%3F-instance_method

PR #336, opened 2018-12-03
Add dependency unixodbc

ofek:
See https://github.com/DataDog/datadog-agent/pull/2729

### DIFF
--- a/config/software/datadog-agent-integrations.rb
+++ b/config/software/datadog-agent-integrations.rb
@@ -4,6 +4,7 @@ name 'datadog-agent-integrations'
 
 dependency 'pip'
 dependency 'datadog-agent'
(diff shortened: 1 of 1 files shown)


config/software/datadog-agent-integrations.rb
```diff
@@ -4,6 +4,7 @@
 
 dependency 'pip'
 dependency 'datadog-agent'
+dependency 'unixodbc'
```
olivielpeau: is this meant to be added to the Windows and macOS builds as well? (https://github.com/DataDog/datadog-agent/pull/2729 only applied to Linux)

ofek:
Superseded by https://github.com/DataDog/dd-agent-omnibus/pull/334

PR #338, opened 2018-12-03
exclude ibm_mq

gmmeyer:
We are only supporting this on agent 6 because of binary dependencies. I am not going to upgrade the build containers to fix this on agent 5.


gmmeyer:
I can confirm that the build works btw

PR #354, opened 2019-02-07
Whitelist A5 integrations

nmuesch:
Used the `ddev` tool to list all integrations that were shipped with Agent 6.9/5.31
Created a whitelist of these integrations and have omnibus only install those integrations

This will allow us to not ship new integrations with Agent 5, but still ship updates and fixes to integrations that have already been shipped with previous versions of A5

Note: This will not stop dependencies from being installed. So new dependencies added to the datadog_checks_base's static dependency list will be included in A5, however their corresponding checks won't be used.


lib/integration_whitelist.rb
```diff
@@ -0,0 +1,113 @@
+INTEGRATION_WHITELIST = [
```
olivielpeau: is the name of the directory in `integrations-core` always the same name as what `ddev` shows?

nmuesch: I just used this function https://github.com/DataDog/integrations-core/blob/d18461375c85522e8bdf87fb67eef7740f151c0f/datadog_checks_dev/datadog_checks/dev/tooling/utils.py#L105 which prints the path of the integration, so the answer should be yes.

nmuesch:
I put some thought into the dependency case and have thought of the following:
1) We create a new integration in core and it includes a new dependency. -> This means the new integration isn't in the whitelist, buts its dep is installed and shipped with A5 because of the reliance on the static deps file - https://github.com/DataDog/dd-agent-omnibus/blob/master/config/software/datadog-agent-integrations.rb#L118. So while the new integration sin't shipped, the related dependency would be. I can't find a good solution to avoid this, but I also can't see it offering any real issues aside from extra weight to the package.

2) An update to an existing integration that is changed because the APi of a dependency changed. Since we still build the dep list and the check from `master` of integrations-core, the compatibility should stay in sync

3) Transitive dependencies are updated which conflict with a direct dependency in the static dep file. This would be caught by the integrations-core tooling and not be merged to master anyway (so an issue outside of this PR that is handled elsewhere)

These are the main cases I could think of and don't think they pose any real issues. Let me know what you think.

lib/integration_whitelist.rb
```diff
@@ -0,0 +1,113 @@
...
+  'nagios',
+  'network',
+  'nfsstat',
+  'nginx',
+  'nginx_ingress_controller',
+  'ntp',
+  'openldap',
+  'openmetrics',
+  'openstack',
+  'openstack_controller',
```
gzussa: Could we remove `openstack_controller`. I believe it never was included at this point, was it?

lib/integration_whitelist.rb
```diff
@@ -0,0 +1,113 @@
...
+  'gitlab',
+  'gitlab_runner',
+  'go-metro',
+  'go_expvar',
+  'gunicorn',
+  'haproxy',
+  'hdfs_datanode',
+  'hdfs_namenode',
+  'http_check',
+  'hyperv',
```
gzussa: Has this been shipped to A5 already? If not let's remove it as well and include it only in A6

lib/integration_whitelist.rb
nmuesch: Good catch. It was included because it was in the 6.9 branch, but I forgot to remove it since it was in the original blacklist.

lib/integration_whitelist.rb
nmuesch: Hyperv is indeed in the 6.9.x branch and wasn't a part of the omnibus blacklist, so it should have been shipped already.

PR #360, opened 2019-03-07
[trace-agent] some omnibus cleanup + build resilience

truthbk:


truthbk:
Found an issue with this, let's not merge yet. Glob matching looks broken.

PR #361, opened 2019-03-15
[protobuf] Remove from omnibus deps of Agent

olivielpeau:
Is installed as binary wheel by integrations software def. See full context in https://github.com/DataDog/datadog-agent/pull/3160


olivielpeau:
Omnibus builds still work on all platforms. Merging

PR #368, opened 2019-04-05
[curl] we don't need curl on windows

truthbk:
pycurl is required, but we have a binary wheel for that. The actual curl library is not required. 

For consistency (although these definitions are actually using the builder provided curl, not the embedded one), let's use the powershell native mechanisms for downloading the deps.


truthbk:
Working build tested successfully! 🚀

PR #377, opened 2019-06-04
Adding constraint for public_suffix

hush-hush:


Gemfile
```diff
@@ -12,3 +12,5 @@ gem 'rake', '~> 11.0'
 
 # We need an older version of mixlib-cli on our version of ruby
 gem 'mixlib-cli', '~> 1.7.0'
+
```
truthbk: nit: I think we can remove this space.

PR #381, opened 2019-10-15
Add 5.32.5-rc.2 version to the release.json

albertvaka:
Everything stays the same but branch `agent-v5` of `omnibus-software` has been updated (which makes me think that we can't have reproducible builds of the previous versions, since we pin a moving branch).


olivielpeau:
(to respond to your comment on not pinning a precise commit of `omnibus-software`, it's true that if we wanted to rebuild it in the future we'd have to change this to the tip of the branch at the time of the original build. Ideally we should tag `omnibus-software` and use that tag, but we should make it so it doesn't require an additional manual step)

PR #401, opened 2021-07-09
[build] addresses build issues - required for centos + windows builds

truthbk:
This PR addresses to separate issues that were causing build failures on centos and windows respectively:
- `pytest-runner` dependency pull triggered by `cisco_aci` was failing, requiring a prior manual install of the dependency.
- `dep` having issue checking out the correct revision for `github.com/openshift/api` on windows due to extraneous characters in `master` filenames (`:` to be specific). The workaround is to clone to branch `release-3.9` which contains the correct expected revision, and `dep` will skip the clone letting us get through.

### DIFF
--- a/omnibus_build.sh
+++ b/omnibus_build.sh
@@ -59,7 +59,7 @@ if [ -n "$LOCAL_AGENT_REPO" ]; then
   cd -
 else
   echo "Querying: $REMOTE_AGENT_REPO_RAW/$AGENT_BRANCH/config.py"
-  curl -v $REMOTE_AGENT_REPO_RAW/$AGENT_BRANCH/config.py 2>/dev/null | grep 'JMX_VERSION'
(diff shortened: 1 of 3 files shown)


omnibus_build.sh
```diff
@@ -58,9 +58,12 @@ if [ -n "$LOCAL_AGENT_REPO" ]; then
   JMX_VERSION=$(git show $AGENT_BRANCH:config.py | grep 'JMX_VERSION' | cut -f2 -d'=' | tr -d ' "')
   cd -
 else
+  echo "Querying: $REMOTE_AGENT_REPO_RAW/$AGENT_BRANCH/config.py"
+  curl -v $REMOTE_AGENT_REPO_RAW/$AGENT_BRANCH/config.py 2>/dev/null | grep 'JMX_VERSION'
```
sgnn7, line 62:
```suggestion
  curl -v "$REMOTE_AGENT_REPO_RAW/$AGENT_BRANCH/config.py" 2>/dev/null | grep 'JMX_VERSION'
```

sgnn7 (approved):
@truthbk One tiny comment but lgtm

